PR #42, opened 2023-07-07
add missing quote in example command - RScript Plugin

npranav10:
When I was looking at the documentation for RScript Plugin, I came across a missing quotation mark at the end of the command (see image below).

image

Assuming the documentation was generated from the file [‎plugin-script-r/src/main/java/io/kestra/plugin/scripts/r/Script.java](https://github.com/kestra-io/plugin-scripts/blob/70cebe911629c9c9dc4156e7b217225bcc4db827/plugin-script-r/src/main/java/io/kestra/plugin/scripts/r/Script.java#L34), I made the required changes in the same file.


tchiotludo:
yes the doc is generated directly from the plugin, thanks :rocket:

PR #46, opened 2023-07-10
fix(deps): force deps version for shadowJar to allow import in other …

brian-mulier-p:
…plugins

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,6 @@ nexusPublishing {
 
 subprojects { Project subproject ->
     apply plugin: "com.adarshr.test-logger"
-    apply plugin: "com.github.johnrengelman.shadow"
     apply plugin: "maven-publish"
     apply plugin: "signing"
     apply plugin: 'ru.vyarus.github-info'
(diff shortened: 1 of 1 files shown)


brian-mulier-p:
I'm not sure about this change...
I think we are in a wrong situation because we are using shadowJar for a library meaning we embark unwanted dependencies without the ability to provide transitive dependencies as-is...
Or we could but it would get verbose because we would have to duplicate the include inside the shadowJar.

The chosen fix here is to force the resolution version but it leads to some cons:
- Idk if dependabot will be able to detect it and ensure we up the version as it goes
- Shadowing dependencies is such a bad practice imo because it could lead to dependencies conflicts which would not be detected by gradle you lose track of which dependency brings what due to shadowJar when you're on the final project which is importing it...

WDYT ?

brian-mulier-p:
Also, another solution would be to exclude dependency conflicts brought by docker-java-client since it brings a lot of old versions which leads to tests failing

loicmathieu (approved):
LGTM

PR #53, opened 2023-07-19
fix(script): No longer splitting too long logs on Docker runner

### LINKED ISSUE #52: Docker runner split logs in multi-line if too long
### Expected Behavior

_No response_

### Actual Behaviour

See https://github.com/docker-java/docker-java/issues/1647 & https://github.com/docker-java/docker-java/pull/1754 & https://github.com/docker-java/docker-java/pull/2050 which have not been addressed (and they probably won't soon).

This is due to the internal engine of java-docker-client which uses a fixed-size buffer preventing logs from being more than 1024 bytes long (else it is splitting it in multi line)
See this screen where my log gets split into multiple lines:
image

This is particularly an issue when you want for eg. to send a token from task to task as they can be really long, causing the echo ::{"outputs" to be split ence not recognized by the current log parser

### Steps To Reproduce

_No response_

### Environment Information

- Kestra Version:
- Plugin version: 
- Operating System (OS / Docker / Kubernetes):
- Java Version (If not docker):

### Example flow

_No response_

brian-mulier-p:
closes #52

### DIFF
--- a/plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/DockerScriptRunner.java
+++ b/plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/DockerScriptRunner.java
@@ -29,15 +29,20 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static io.kestra.core.utils.Rethrow.throwConsumer;
 import static io.kestra.core.utils.Rethrow.throwFunction;
 
 public class DockerScriptRunner {
     private static final ReadableBytesTypeConverter READABLE_BYTES_TYPE_CONVERTER = new ReadableBytesTypeConverter();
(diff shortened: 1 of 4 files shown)


brian-mulier-p:
Putting it as a draft since I'm not sure about my implementation, it feels strange to include the Kestra expression syntax (::{ & }::) but this is the only type of log line which we can be sure should be grouped together.

Since docker-java-client doesn't provide any way to know that a log is only part of a line, I don't see any other way to get around

plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/DockerScriptRunner.java
```diff
@@ -128,17 +132,37 @@ public RunnerResult run(CommandsWrapper commands, DockerOptions dockerOptions) t
...
                         @Override
-                        public void onNext(Frame item) {
-                            defaultLogConsumer.accept(
-                                new String(item.getPayload()).trim(),
-                                item.getStreamType() == StreamType.STDERR
-                            );
+                        public void onNext(Frame frame) {
+                            String frameStr = new String(frame.getPayload());
+
+                            Pattern pattern = Pattern.compile("[\\r\\n]+$");
```
loicmathieu, line 142: Better to declare the pattern as a constant to avoid compiling again and again

loicmathieu (approved):
LGTM if you check on real use case with concurrent bash to be sure that logs are not mixed up together as this is not thread safe

Issue #51, opened 2023-07-19
OutputDir examples


Skraye:
### Issue description

Write an example for each task including the usage of OutputDir.

PR #55, opened 2023-08-04
(docs) outputDir and more examples

anna-geller:
closes https://github.com/kestra-io/plugin-scripts/issues/51

closes https://github.com/kestra-io/kestra.io/issues/367 (we can add more documentation later on if we see demand for it)

### DIFF
--- a/plugin-script-r/src/main/java/io/kestra/plugin/scripts/r/Script.java
+++ b/plugin-script-r/src/main/java/io/kestra/plugin/scripts/r/Script.java
@@ -39,6 +39,48 @@ import java.util.List;
                 "  - Rscript -e 'install.packages(\"lubridate\")'"
             }
         ),
+        @Example(
+            full = true,
+            title = """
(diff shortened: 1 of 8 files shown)


plugin-script-r/src/main/java/io/kestra/plugin/scripts/r/Script.java
```diff
@@ -39,6 +39,48 @@
...
+            code = """
+                id: rCars
+                namespace: dev
+
+                tasks:
+                  - id: r
+                    type: io.kestra.plugin.scripts.r.Script
+                    warningOnStdErr: false
+                    docker:
+                    image: ghcr.io/kestra-io/rdata:latest
```
loicmathieu, line 56:
```suggestion
                      image: ghcr.io/kestra-io/rdata:latest
```
Wrong YAML indentation

loicmathieu (approved):
Nice exemples!

I spotted one wrongly indented YAML but I didn't look into details on all examples.

As a side note, in Java, text blocks (aka multi-line strings) are usually aligned with the enclosing `"""`, you did this for some but not all. You can merge as is if you want as it will not make any difference on the generated example.

PR #58, opened 2023-08-10
feat(tasks): introduce old script emulation

tchiotludo:


plugin-script/src/main/java/io/kestra/core/tasks/scripts/AbstractPython.java
```diff
@@ -0,0 +1,28 @@
...
+
+@SuperBuilder
+@ToString
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+@Slf4j
+@Deprecated
+public abstract class AbstractPython extends AbstractBash {
+
```
loicmathieu, line 27: It seems odd to me that this task didn't contains the previous properties it has on the core: https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/tasks/scripts/AbstractPython.java

loicmathieu (changes requested):
I review this PR by assuming you copy/paste the code from core and only modify the AbstractBash class. If it's the case it's good for me except the empty abstract class AbstractPython that looks suspicious.

if you update other part of the code, please tell me so I can have a look in details on those areas.

plugin-script/src/main/java/io/kestra/core/tasks/scripts/AbstractPython.java
tchiotludo: You are true, the AbstractPython is useless since there props are copied on the Python, I have remove then

loicmathieu (approved):
LGTM

PR #79, opened 2023-10-13
fix: do not match end of line only in line ending

moadibfr:
some tools use \r that will appear multuple time in the same frame so we should handle that otherwise logs are only flushed in the end.

### DIFF
--- a/plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/DockerScriptRunner.java
+++ b/plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/DockerScriptRunner.java
@@ -45,7 +45,7 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
 
 public class DockerScriptRunner {
     private static final ReadableBytesTypeConverter READABLE_BYTES_TYPE_CONVERTER = new ReadableBytesTypeConverter();
-    public static final Pattern NEWLINE_PATTERN = Pattern.compile("[\\r\\n]+$");
+    public static final Pattern NEWLINE_PATTERN = Pattern.compile("([^\\r\\n]+)[\\r\\n]+");
 
     private final RetryUtils retryUtils;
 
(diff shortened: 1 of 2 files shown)


plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/DockerScriptRunner.java
```diff
@@ -45,7 +45,7 @@
 
 public class DockerScriptRunner {
     private static final ReadableBytesTypeConverter READABLE_BYTES_TYPE_CONVERTER = new ReadableBytesTypeConverter();
-    public static final Pattern NEWLINE_PATTERN = Pattern.compile("[\\r\\n]+$");
+    public static final Pattern NEWLINE_PATTERN = Pattern.compile("([^\\r\\n]+)[\\r\\n]+");
```
Skraye, line 48: Isnt this one unnecessary as it catch the same as before, and the previous one is already annotated with `+` ? 
```suggestion
    public static final Pattern NEWLINE_PATTERN = Pattern.compile("([^\\r\\n]+)");
```

moadibfr: nop this does not work we explicitely want the pattern to end with one of these char.
Like if the frame we get contains no newline it would match with your pattern but should not be sent:
```
Log("This log should not be flushed until we reach the newline or carry return here: \r")

onNext("his log should not be flushed until we reach ") < This would flush as it match the pattern without termination (basically it would match every frame at least once)
```

moadibfr: we don't want to catch the end of the frame if it does not end with \r or \n. If there is a last line without line endin the oncomplete will flush it

Skraye (approved):
LGTM

PR #92, opened 2024-01-04
fix(script): no Docker options when not using the Docker runner

yuri1969:

This prevents using incompatible settings with the `PROCESS` runner.

Note, merging this would bring a breaking change. There might be a better way around this.

close kestra-io/kestra/issues/2132

---

```yaml
id: hello-world-docker-options
namespace: company.team
tasks:
  - id: docker-only
    type: io.kestra.plugin.scripts.shell.Commands
    runner: PROCESS # FIXME
    docker:
      image: foo
    commands:
      - echo 'bar'
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,6 +92,7 @@ subprojects {
 
         testImplementation platform("io.micronaut.platform:micronaut-platform:$micronautVersion")
         testImplementation "io.micronaut.test:micronaut-test-junit5"
+        testImplementation "io.micronaut.validation:micronaut-validation"
         testImplementation "io.micronaut.reactor:micronaut-reactor"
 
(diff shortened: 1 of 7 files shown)


loicmathieu:
Hi,
So we decided to postpone a little as we're not sure if we want it to fail the validation or just add a warning, which is not yet possible with our current validation implementation.

yuri1969:
Understood, having OK/warning/error validation states available would be useful.

anna-geller:
closing as it will no longer be an issue with Task Runners https://develop.kestra.io/docs/concepts/task-runners

`docker` property will no longer be available when using the Process task runner

Thanks so much for flagging this issue @yuri1969, it helped to shape the implementation of task runners

PR #99, opened 2024-01-25
fix: suppress Python warning

anna-geller:
Currently, whenever you install any packages using `beforeCommands`, you'll always see this annoying warning: `WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behavior with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv`

it can be easily suppressed with an environment variable

### DIFF
--- a/plugin-script-node/src/main/java/io/kestra/plugin/scripts/node/Script.java
+++ b/plugin-script-node/src/main/java/io/kestra/plugin/scripts/node/Script.java
@@ -150,7 +150,10 @@ public class Script extends AbstractExecScript {
         );
 
         return commands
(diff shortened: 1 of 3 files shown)


tchiotludo:
there is missing place, look at this:
image

at least one other on these project

anna-geller:
for node, I believe the pip warning is not needed as it uses npm (not pip) but I still added it + for the deprecated core tasks, we can skip

for all python-based I added them:
- https://github.com/kestra-io/plugin-dbt/pull/83
- https://github.com/kestra-io/plugin-singer/pull/50
- https://github.com/kestra-io/plugin-soda/pull/34

PR #123, opened 2024-03-19
feat: script runner

loicmathieu:
Allow to run script in a ScriptRunner provided by a plugin.


loicmathieu:
@anna-geller, there is a little breaking change introduced here, we align all container labels.

Previous labels were:
```
flow.kestra.io/id
flow.kestra.io/namespace
task.kestra.io/id
execution.kestra.io/id
taskrun.kestra.io/id
```

New labels are:
```
kestra.io/namespace
kestra.io/flow-id
kestra.io/task-id
kestra.io/execution-id
kestra.io/taskrun-id
kestra.io/taskrun-attempt
```

anna-geller:
why would autoassigned labels be considered a breaking change?

loicmathieu:
If people have automation based on that, there is a very small risk so it's up to you to note that in the release note or not.

PR #124, opened 2024-03-22
fix: change to a default image that has kestra libs included

anna-geller:
https://github.com/kestra-io/examples/blob/main/dockerfiles/kestrapy.Dockerfile


loicmathieu (approved):
LGTM

PR #130, opened 2024-04-03
feat/script-runners-for-scripts

brian-mulier-p:


build.gradle
```diff
@@ -22,6 +22,9 @@ group "io.kestra.plugin"
 
 subprojects {
     group "io.kestra.plugin"
+
+    apply plugin: "java-library"
+
```
loicmathieu, line 27: I'm not sure it's correct to move it there

loicmathieu (commented):
The main issue is that you had to update all existing Script task which means if there is any in another plugin it also needs to be done.
But I agree we are at a point where we didn't have choice.

The Malloy plugin also needs to be updated

build.gradle
brian-mulier-p: I was unable to build plugin-script as it needed to apply java plugin before specifying the target / source compatibility (I had error where it said that dependency script is not built for proper java version)

plugin-script/src/main/java/io/kestra/plugin/scripts/exec/scripts/services/ScriptService.java
```diff
@@ -18,11 +20,16 @@ private ScriptService() {
...
 
     public static List<String> uploadInputFiles(RunContext runContext, List<String> commands) throws IOException {
-        return io.kestra.core.models.script.ScriptService.uploadInputFiles(runContext, commands);
+        try {
+            return io.kestra.core.models.script.ScriptService.replaceInternalStorage(runContext, Collections.emptyMap(), commands, (ignored, file) -> {}, false);
+        } catch (IllegalVariableEvaluationException e) {
+            // Throw unchecked exception to prevent breaking the old method signature
+            throw new RuntimeException(e);
+        }
     }
```
loicmathieu, line 33: Not mandatory but better to create a new method that declare throwing this exception then make the existing deprecated and using the new method

loicmathieu (approved):
LGTM

PR #140, opened 2024-04-22
refactor: migrate to Java ServiceLoader for plugin registration

fhussonnois:


plugin-script/src/main/java/io/kestra/plugin/scripts/exec/AbstractExecScript.java
```diff
@@ -33,15 +32,14 @@ public abstract class AbstractExecScript extends Task implements RunnableTask<Sc
...
-    @NotEmpty
+    @NotNull
     protected RunnerType runner = RunnerType.DOCKER;
 
     @Schema(
         title = "The task runner to use.",
         description = "Task runners are provided by plugins, each have their own properties."
     )
     @PluginProperty
-    @Valid
```
loicmathieu, line 44: AFAIK it should be kept or validation will not be done inside the TaskRunner attribute itself

plugin-script/src/main/java/io/kestra/plugin/scripts/runner/docker/DockerTaskRunner.java
```diff
@@ -100,6 +99,7 @@
     },
     beta = true // all task runners are beta for now, but this one is stable as it was the one used before
 )
+@JsonDeserialize
```
loicmathieu, line 102: IS it really needed?

PR #168, opened 2024-08-19
feat: JBang scripts

loicmathieu:
Fixes https://github.com/kestra-io/kestra/issues/2150


anna-geller:
Can you change plugin examples to be full examples? (The team has recently agreed to always have full examples). I think for the wording, we may rewrite to "Execute a script written in Java" instead of "Execute a Java script" to avoid confusion with JS

PR #177, opened 2024-09-23
feat(docs): add examples for include/exclude in namespace files

shrutimantri:


plugin-script-shell/src/main/java/io/kestra/plugin/scripts/shell/Commands.java
```diff
@@ -56,6 +56,46 @@
...
+            title = "Exclude specific namespace files.",
+            full = true,
+            code = """
+                   id: exclude_files
+                   namespace: company.team
+
+                   tasks:
+                     - id: command
+                       type: io.kestra.plugin.scripts.shell.Commands
+                       description: "All the `namespaceFiles` except those that are excluded get listed"
```
anna-geller, line 89:
```suggestion
                       description: "All `namespaceFiles` except those that are excluded will be injected into the task's working directory"
```

anna-geller (approved):
LGTM, ty!

PR #185, opened 2024-11-06
Doc fix for Python script task - {{ outputDir }} is deprecated

camsellem:

Updated the doc for the Python task since {{ outputDir }} is now deprecated.

Examples have been tested within Kestra. Flows are running successfully.

Note that this is not a replacement for unit tests but rather a way to demonstrate how the changes work in a real-life scenario, as the end-user would experience them.

Remove this section if this change applies to all flows or to the documentation only. -->

---

Thank you for your contribution. ❤️  -->


anna-geller (approved):
Thanks a lot! For posterity, it's related to this deprecation: https://kestra.io/docs/migration-guide/0.17.0/local-files

PR #213, opened 2025-02-10
fix: rollback commands to PluginProperty to permit workingDir rendering

mgabelle:
fix: rollback commands to PluginProperty to permit workingDir rendering

WorkingDir is rendered inside Kestra. 
It is causing issue when present in `commands` or `script` because with Property we render them before going into kestra and they don't have access to the workingDir var.


loicmathieu:
I'm not sure this was the right fix.
Another way of doing this was to pass the Property<Script> to the script service so the rendering would be done later as before.

PR #240, opened 2025-04-29
feat(python): add support for package management

fhussonnois:
First implementation to add support for installing and managing declared Python packages

Add new task properties `dependencies`, `pythonVersion` and `dependencyCacheEnabled`.

Related-to: kestra-io/kestra-ee#3530

### DIFF
--- a/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
+++ b/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
@@ -1,0 +1,156 @@
+package io.kestra.plugin.scripts.python;
+
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.models.executions.metrics.Timer;
+import io.kestra.core.models.property.Property;
+import io.kestra.core.runners.DefaultRunContext;
+import io.kestra.core.runners.RunContext;
+import io.kestra.plugin.core.runner.Process;
+import io.kestra.plugin.scripts.exec.AbstractExecScript;
+import io.kestra.plugin.scripts.exec.scripts.models.RunnerType;
+import io.kestra.plugin.scripts.python.internals.PythonDependenciesResolver;
+import io.kestra.plugin.scripts.python.internals.PythonVersionParser;
+import io.kestra.plugin.scripts.python.internals.ResolvedPythonPackages;
+import io.swagger.v3.oas.annotations.media.Schema;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Path;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+@SuperBuilder
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+public abstract class AbstractPythonExecScript extends AbstractExecScript {
+
+    public static final String DEFAULT_PYTHON_VERSION = "3.13";
+    public static final String DEFAULT_IMAGE = "python:" + DEFAULT_PYTHON_VERSION + "-slim";
+    public static final List<String> DEFAULT_PYTHON_PACKAGES = List.of("kestra", "amazon-ion");
+
+    @Builder.Default
+    protected Property<String> containerImage = Property.of(DEFAULT_IMAGE);
+
+    @Schema(
+        title = "The script dependencies."
+    )
+    @PluginProperty
+    protected Property<List<String>> dependencies;
+
+    @Schema(
+        title = "The version of Python to use for the script.",
+        description = "If no version is explicitly specified, the task will attempt to extract the version from the configured container image. If it cannot determine the version from the image, the task will default to Python '"+ DEFAULT_PYTHON_VERSION +" '"
+    )
+    @PluginProperty
+    protected Property<String> pythonVersion;
+
+    @Schema(
+        title = "Enable Python dependency caching",
+        description = "When enabled, Python dependencies will be cached across task executions. This locks dependency versions and speeds up subsequent runs by avoiding redundant installations."
+    )
+    @PluginProperty
+    protected Property<Boolean> dependencyCacheEnabled = Property.of(true);
+
+    protected ResolvedPythonEnvironment setupPythonEnvironment(final RunContext runContext, final boolean isCacheEnabled) throws IllegalVariableEvaluationException, IOException {
+        List<String> requirements = new ArrayList<>(runContext.render(dependencies).asList(String.class));
+        DEFAULT_PYTHON_PACKAGES.forEach(pkg -> addPackageIfNoneMatch(requirements, pkg));
+
+        final String targetPythonVersion = getTargetPythonVersion(runContext);
+
+        final Path localCacheDir = getLocalCacheDir(runContext);
+        final PythonDependenciesResolver resolver = new PythonDependenciesResolver(runContext.logger(), runContext.workingDir(), localCacheDir);
+
+        final String hash = resolver.getRequirementsHashKey(targetPythonVersion, requirements);
+
+        final long metricCacheDownloadStart = System.currentTimeMillis();
+
+        Optional<InputStream> cacheFile = isCacheEnabled ? runContext.storage().getCacheFile("python-" + getType(), hash) : Optional.empty();
+
+        final ResolvedPythonPackages resolvedPythonPackages;
+        final boolean cached;
+        if (cacheFile.isPresent()) {
+            runContext.logger().debug("Restoring python dependencies cache for key: {}", hash);
+            resolvedPythonPackages = resolver.getPythonLibs(targetPythonVersion, hash, cacheFile.get());
+            runContext.logger().debug("Cache restored successfully");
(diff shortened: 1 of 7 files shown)


brian-mulier-p (approved):
Not much, LGTM if you tackle / think my comments are invalid around per-dep cache

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Schema(
+        title = "The script dependencies."
+    )
+    @PluginProperty
+    protected Property<List<String>> dependencies;
+
+    @Schema(
+        title = "The version of Python to use for the script.",
+        description = "If no version is explicitly specified, the task will attempt to extract the version from the configured container image. If it cannot determine the version from the image, the task will default to Python '"+ DEFAULT_IMAGE +" '"
```
loicmathieu, line 52:
```suggestion
        description = "If no version is explicitly specified, the task will attempt to extract the version from the configured container image. If it cannot determine the version from the image, the task will default to Python '"+ DEFAULT_PYTHON_VERSION +" '"
```

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
```diff
@@ -0,0 +1,149 @@
...
+
+    protected ResolvedPythonEnvironment setupPythonEnvironment(final RunContext runContext, final boolean isCacheEnabled) throws IllegalVariableEvaluationException, IOException {
+        List<String> requirements = new ArrayList<>(runContext.render(dependencies).asList(String.class));
+        DEFAULT_PYTHON_PACKAGES.forEach(pkg -> addPackageIfNoneMatch(requirements, pkg));
+
+        final String targetPythonVersion = getTargetPythonVersion(runContext);
+
+        final Path localCacheDir = ((DefaultRunContext) runContext).getApplicationContext().getEnvironment().getProperty("kestra.cache.dir", String.class)
+            .map(Path::of)
+            .orElse(Path.of("/tmp/kestra/cache"));
```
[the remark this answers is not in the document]
loicmathieu: There is already a kestra temp dir config.
The local cache should be inside it I think.

`kestra.tasks.tmp-dir.path`.

Anyway, you don't want to hardcode /tmp so better use the `java.io.tmpdir` sys pro

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
```diff
@@ -0,0 +1,149 @@
...
+
+        Optional<InputStream> cacheFile = isCacheEnabled ? runContext.storage().getCacheFile("python-" + getType(), hash) : Optional.empty();
+
+        final ResolvedPythonPackages resolvedPythonPackages;
+        final boolean cached;
+        if (cacheFile.isPresent()) {
+            runContext.logger().debug("Restoring python dependencies cache for key: {}", hash);
+            resolvedPythonPackages = resolver.getPythonLibs(targetPythonVersion, hash, cacheFile.get());
+            runContext.logger().debug("Cache restored successfully");
+            runContext.metric(Timer.of("cache.download.duration", Duration.ofMillis(System.currentTimeMillis() - metricCacheDownloadStart)));
```
loicmathieu, line 88: metric names as unique in the whole Kestra instance.
Maybe use a more specific one like dependency.cache.download.duration?
Or even starting with task or plugin?

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
```diff
@@ -0,0 +1,149 @@
...
+            if (isCacheEnabled) {
+                runContext.logger().debug("Could not find python dependencies cache for key: {}", hash);
+            }
+            resolvedPythonPackages = resolver.getPythonLibs(targetPythonVersion, hash, requirements);
+            cached = false;
+        }
+
+        runContext.logger().debug("Installed dependencies: {}", resolvedPythonPackages.packagesToString());
+
+        String pythonInterpreter = "python";
```
loicmathieu, line 100: The interpreter can be configured in the Python task itself so better use it from there.
Example usage: python3 instead of python :(

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
```diff
@@ -0,0 +1,149 @@
...
+        return new ResolvedPythonEnvironment(cached, resolvedPythonPackages, pythonInterpreter);
+    }
+
+    protected void uploadCache(final RunContext runContext, final ResolvedPythonPackages resolvedPythonPackages) throws IOException {
+        final long start = System.currentTimeMillis();
+        runContext.logger().debug("Uploading python dependencies cache for key: {}", resolvedPythonPackages.hash());
+        File cache = resolvedPythonPackages.toZippedArchive(runContext.workingDir());
+        runContext.storage().putCacheFile(cache, "python-" + getType(), resolvedPythonPackages.hash());
+        runContext.logger().debug("Cache uploaded successfully (size: {} bytes)", cache.length());
+        runContext.metric(Timer.of("cache.upload.duration", Duration.ofMillis(System.currentTimeMillis() - start)));
```
loicmathieu, line 114: same remark for the metric name

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Commands.java
```diff
@@ -299,17 +294,26 @@ protected DockerOptions injectDefaults(RunContext runContext, DockerOptions orig
...
+        boolean isCacheEnabled = runContext.render(this.dependencyCacheEnabled).as(Boolean.class).orElse(true);
+        ResolvedPythonEnvironment pythonEnvironment = setupPythonEnvironment(runContext, isCacheEnabled);
+
+        ScriptOutput output = this.commands(runContext)
             .addEnv(Map.of(
                 "PYTHONUNBUFFERED", "true",
                 "PIP_ROOT_USER_ACTION", "ignore",
-                "PIP_DISABLE_PIP_VERSION_CHECK", "1"
+                "PIP_DISABLE_PIP_VERSION_CHECK", "1",
+                "PYTHONPATH", pythonEnvironment.packages().path().toString()
```
loicmathieu, line 305: What if pip or uv is used as a before command?
Would it work like before if you set the PYTHONPATH?

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Commands.java
```diff
@@ -299,17 +294,26 @@ protected DockerOptions injectDefaults(RunContext runContext, DockerOptions orig
...
             ))
             .withInterpreter(this.interpreter)
             .withCommands(commands)
             .withBeforeCommands(beforeCommands)
             .withBeforeCommandsWithOptions(true)
             .withTargetOS(os)
             .run();
+
+        if (output.getExitCode() == 0 && !pythonEnvironment.cached() && isCacheEnabled) {
```
loicmathieu, line 314: Caching should be independant of whether or not the script fail

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Script.java
```diff
@@ -324,20 +318,28 @@ public ScriptOutput run(RunContext runContext) throws Exception {
...
             .withBeforeCommands(beforeCommands)
             .withBeforeCommandsWithOptions(true)
             .withCommands(Property.of(List.of(
-                String.join(" ", "python", commands.getTaskRunner().toAbsolutePath(runContext, commands, relativeScriptPath.toString(), os))
+                String.join(" ", pythonEnvironment.interpreter(), commands.getTaskRunner().toAbsolutePath(runContext, commands, relativeScriptPath.toString(), os))
             )))
             .withTargetOS(os)
             .run();
+
+        if (output.getExitCode() == 0 && !pythonEnvironment.cached() && isCacheEnabled) {
```
loicmathieu, line 340: Same here, I'll upload the cache even if the script fail

loicmathieu (approved):
Overall it's a very nice addition, I think it can be abstracted later to be use for other scripts that manage dependencies (NodeJS, maybe JBang)

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
fhussonnois: Actually, the  `python` interepreter is already hard-coded right in 0.22x version. It's only on the legacy Python Task that the user could configure it. So this PR doesn't really change the behavior.

plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/AbstractPythonExecScript.java
```diff
@@ -0,0 +1,149 @@
...
+
+        final String targetPythonVersion = getTargetPythonVersion(runContext);
+
+        final Path localCacheDir = ((DefaultRunContext) runContext).getApplicationContext().getEnvironment().getProperty("kestra.cache.dir", String.class)
+            .map(Path::of)
+            .orElse(Path.of("/tmp/kestra/cache"));
+
+        final PythonDependenciesResolver resolver = new PythonDependenciesResolver(runContext.logger(), runContext.workingDir(), localCacheDir);
+
+        final String hash = resolver.getRequirementsHashKey(targetPythonVersion, requirements);
```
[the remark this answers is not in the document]
fhussonnois: Basically, the cache is created for the entire `requirement.txt` and a python version. And because, in all cases, the cache is sticked to the flow (we may change that later) there is no value to manage dependencies independantly - it would be counterproductive (retrieving a cache stay costly) - Also note that we adopt here a strategy already used by some Kestra alternatives.

PR #242, opened 2025-05-11
feat(windows-plugin): add cmd.exe-based plugin for executing .exe fil…

balaaparna-dev:
Closes https://github.com/kestra-io/kestra/issues/2510.

To Enable Windows Command Execution

The task supports use cases such as:
- Running .exe files.
- Interacting with the local file system .
related to issue #2510.
---

Two unit tests were implemented in the WindowsCmdTaskTest class to validate the functionality of the WindowsCmdTask implementation:
testSimpleCommand: Verifies that a valid command (echo Hello, World!) executes successfully and produces the expected output.
testInvalidCommand: Ensures that an invalid command (invalidcommand) is handled gracefully by throwing an appropriate exception.


MilosPaunovic:
Just a small note for the future, if you could use **GitHub Keywords** when opening PRs to connect it to the proper issue (to automate the entire process a bit more), it would be super helpful to us. You can read a bit more about it [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) (I've added closing keyword for this particular issue).

plugin-script-windows/src/main/java/io/kestra/plugin/scripts/windows/WindowsCmdTask.java
```diff
@@ -0,0 +1,158 @@
...
+            code = {
+                "command: \"dir\""
+            }
+        ),
+        @Example(
+            title = "Run an executable file",
+            code = {
+                "command: \"C:\\\\path\\\\to\\\\program.exe\""
+            }
+        )
```
fdelbrayelle, line 47: For each example, you need to provide a full flow like this:

```yaml
    @Example(
        full = true,
        title = "Execute Windows commands.",
        code = """
            id: execute_windows_commands
            namespace: company.team

            tasks:
              - id: windows
                type: io.kestra.plugin.scripts.windows.Commands
                commands:
                  - dir
            """
    )
```

plugin-script-windows/src/main/java/io/kestra/plugin/scripts/windows/WindowsCmdTask.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * The Windows command to execute using cmd.exe.
+     */
+    @Schema(
+        title = "Command to execute",
+        description = "The Windows command to execute using cmd.exe."
+    )
+    @PluginProperty(dynamic = true)
+    private String command;
```
fdelbrayelle, line 67: The annotation `@PluginProperty` is the old way to do. Can you please for all properties do like this?

```suggestion
    private Property<String> command;
```

plugin-script-windows/src/main/resources/plugin.yaml
```diff
@@ -0,0 +1,5 @@
+tasks:
+  - id: windows-cmd-task
+    name: Windows Command Task
+    description: Execute Windows commands using cmd.exe
+    class: io.kestra.plugin.scripts.windows.WindowsCmdTask
```
fdelbrayelle, line 5: Where is it used?

anna-geller:
@Ben8t @aj-emerich FYI the planned syntax for this plugin is below (confirmed with @fdelbrayelle):

```yaml
id: windows
namespace: company.team

tasks:
  - id: cli
    type: io.kestra.plugin.scripts.windows.Commands
    commands:
      - dir

  - id: script
    type: io.kestra.plugin.scripts.windows.Script
    script: |
      @echo off
      cls
      rem This is a comment
      echo Hello World!
      pause     
```

fdelbrayelle:
@balaaparna-dev I'm closing the PR. Here is why : the Docker image was missing so your code couldn't never work because "cmd.exe" cannot be executed on a Linux.

I searched for solutions (like using MS-DOS images but they need docker login and it's just emulation because we **cannot** run Windows commands like "cmd.exe" on Linux systems). I think it's overdated and the solution to use is the Powershell script instead actually.

PR #272, opened 2025-07-03
docs(python-scripts): update examples to use dependencies

aj-emerich:


plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Script.java
```diff
@@ -68,8 +69,9 @@
                 tasks:
                   - id: run_python
                     type: io.kestra.plugin.scripts.python.Script
-                    beforeCommands:
-                      - pip install requests kestra
+                    dependencies:
+                      - requests
+                      - kestra
```
wrussell1999, line 74: This flow doesn't use Kestra
```suggestion
```

wrussell1999 (approved):

PR #373, opened 2026-06-18
fix(python): remap PYTHONPATH to the task runner working dir

### LINKED ISSUE #372: Python dependencies not found on the Kubernetes task runner (ModuleNotFoundError)
When a Python `Script`/`Commands` task declares `dependencies` and runs on a remote task runner (e.g. Kubernetes), the script fails with `ModuleNotFoundError`.

The resolved packages are installed under the working dir (`.kestra_additional_python_lib`) and exposed via `PYTHONPATH`, but `PYTHONPATH` is set to the absolute **worker** path (e.g. `/tmp/<id>/.kestra_additional_python_lib`) instead of being remapped to the runner working dir (e.g. `/kestra/working-dir/...`). With the Docker runner the worker working dir is mounted at the same path so it worked; the Kubernetes runner mounts it at `/kestra/working-dir`, so the path doesn't exist in the pod.

Affects `python.Script` and `python.Commands`.

Skraye:
Python `dependencies` were exposed via `PYTHONPATH` using the absolute worker path, so they were missing on remote task runners (Kubernetes) where the working dir is mounted elsewhere → `ModuleNotFoundError`. Remap the packages path through `taskRunner.toAbsolutePath()`, like the script path already is. Applies to both `Script` and `Commands`.

Closes #372

### DIFF
--- a/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Commands.java
+++ b/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Commands.java
@@ -1,5 +1,6 @@
 package io.kestra.plugin.scripts.python;
 
+import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


fdelbrayelle (changes requested):
@Skraye Could you update unit tests accordingly 🙏 ?

Skraye:
@fdelbrayelle We don't have anything that test through K8S
How would you like me to test it?

fdelbrayelle:
I mean here it's not K8S it's Script & Commands on plugin-scripts where you changed code but fair enough LGTM :D

fdelbrayelle (approved):

Skraye:
@fdelbrayelle  the shouldExecScriptGivenDependency test already covers dependencies with the docker and process runners. The bug never showed there because on both the worker and runner working dir are the same path so the missing remap is invisible.

 It only happens on Kubernetes runner, where the working dir is a different path (/kestra/working-dir), so Docker/Process behaviour is unchanged. 

So if we want to add a test that could reproduce the issue without the fix, we would need to add test with K8s runner. I can open an issue for that, probably a cooldown task.

PR #394, opened 2026-07-07
fix(python): resolve real interpreter for Process runner without pythonVersion

jymaire:
### What changes are being made and why?

Python `Script`/`Commands` tasks running on the **`Process` task runner** failed with `/bin/sh: python: not found` (exit code 127) on hosts that ship only `python3` (Debian/Ubuntu, and the standard Kestra worker image).

**Root cause:** In `PythonEnvironmentManager.setup()`, the interpreter was only resolved to an actual installed binary when `pythonVersion` was explicitly set. When it was unset (the common case) on a Process runner, the interpreter stayed the literal `"python"`, which doesn't exist on `python3`-only hosts. The Docker default runner was unaffected because `python:3.13-slim` symlinks `python`.

**Fix:** The Process runner now always resolves an installed interpreter. When `pythonVersion` is set, behavior is unchanged (resolved through the configured package manager, which may provision a managed Python via `uv`). When it's unset, we probe for an already-installed interpreter (`python3` → `python`) via `PackageManagerType.PIP.getPythonPath`, avoiding a managed-Python download just to run a dependency-free script. The Docker/Kubernetes path is untouched.

Also:
- Updated the `all_python` sanity-check flow: `python_commands_process` now uses `python3 main.py` (user-authored command, not routed through the resolver).
- Added `PythonEnvironmentManagerTest` regression test asserting a Process-runner Script with no `pythonVersion` resolves to a runnable, non-`"python"` interpreter.

---

### How the changes have been QAed?

- New `PythonEnvironmentManagerTest` passes.
- `ScriptTest.task[PROCESS]` — the exact case that reproduced the reported failure — now passes (previously exit 127).

```yaml
id: repro
namespace: sanitychecks.plugin-scripts
tasks:
  - id: python_script_process
    type: io.kestra.plugin.scripts.python.Script
    taskRunner:
      type: io.kestra.plugin.core.runner.Process
    script: |
      print("Hello from a Process Task Runner")
```
On a `python3`-only host this failed with `python: not found` before the fix and now succeeds.

---

### Contributor Checklist ✅

- [x] I have read and followed the [plugin contribution guidelines](https://kestra.io/docs/plugin-developer-guide/contribution-guidelines)

PR #399, opened 2026-07-20
fix(python): resolve interpreter through package manager when dependencies are installed

### LINKED ISSUE #397: Python Script: Process runner uses bare 'python' instead of uv-managed interpreter path
## Summary

`io.kestra.plugin.scripts.python.Script`/`Commands` tasks running on the **Process** task runner fail with `python: not found` (exit 127) whenever dependencies are declared, `pythonVersion` is left unset, and the worker host has no `python`/`python3` binary already on `PATH` (uv must download a managed interpreter). The dependency install succeeds against the managed interpreter, but the script is then executed with the bare `python` literal instead of the managed interpreter's path.

## Actual Behaviour

With `dependencies` set and `pythonVersion` unset, on a host with no local Python:

1. `uv python find --system --no-managed-python` → `No interpreter found in search path`.
2. Kestra logs `No Python Version found. Using default version: '3.13'` and lets `uv` provision a managed interpreter at `/tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13`.
3. Dependencies compile/install successfully against that managed interpreter (`Installed dependencies: ...`).
4. The script is executed via `/bin/sh -c "set -e\npython /kestra/working-dir/xxx.py"` — the bare `python` literal, not the managed interpreter path from step 2.
5. Execution fails: `/bin/sh: 2: python: not found`, exit code 127, and the task's outputs can no longer be rendered.

Root cause appears to be in `PythonEnvironmentManager.setup()` (`plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/internals/PythonEnvironmentManager.java`, ~line 106-117): when `taskRunner instanceof Process` and the user-supplied `pythonVersion` is `null`, the resolved interpreter is looked up independently via `PackageManagerType.PIP.getPythonPath(resolver, targetPythonVersion)` instead of reusing the interpreter path that `uv` already resolved/managed while installing the dependencies a few lines above. On a host with no pre-installed Python, that independent lookup finds nothing and falls back to the literal `"python"`.

This is a regression surfaced after #394 (v1.9.5), which fixed the *no-dependency* case (probing `python3`/`python` on `PATH`) but did not align the *with-dependencies-and-no-local-python* case with the interpreter that `uv` actually provisioned.

## Expected Behaviour

When dependencies are declared and no local interpreter exists, the script must be executed with the same managed interpreter path that `uv` resolved and used to install the dependencies (e.g. `/tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13`), not the bare `python` literal.

## Reproducer

Process task runner, no local Python on the worker host (only `uv`-managed), `pythonVersion` left unset:

```yaml
id: python-interpreter-repro
namespace: company.team

tasks:
  - id: exec
    taskRunner:
    dependencies: [cryptography, kestra, lxml, pysaml2, signxml]
    script: |
      if __name__ == "__main__":
          Kestra.outputs({'assertion': 'foobar'})

outputs:
  - id: assertion
    type: STRING
    value: "{{ outputs.exec.vars.assertion }}"
```

## Logs / Stack Trace

```
DEBUG Executing command: uv python find --system --no-managed-python
DEBUG error: No interpreter found in search path
WARN  No Python Version found. Using default version: '3.13'
DEBUG Executing command: uv python find 3.13 --system --python-preference=only-managed
DEBUG /tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13
DEBUG Installing packages with uv
DEBUG Installed dependencies: amazon-ion==0.14.6, certifi==2026.6.17, cffi==2.1.0, ... signxml==5.1.0
DEBUG Starting command with pid 328 [/bin/sh -c set -e
python /tmp/hgEAa2Be7zskjD37DHLx3/18108143320553494109.py]
ERROR /bin/sh: 2: python: not found
ERROR io.kestra.core.models.tasks.runners.TaskException: Command failed with exit code 127
ERROR Failed to render output values: Unable to find `assertion` used in the expression `{{outputs.exec.vars.assertion}}`
```

## Environment

- **Kestra version**: EE v1.0.51 (dev)
- **Plugin version**: plugin-scripts v1.9.5+ (regression persists after the #394 fix)
- **Deployment**: EE self-hosted

## Additional Context

Worked with a previous plugin-scripts version; broke after an update. The same flow, unchanged, runs fine when `pythonVersion` is explicitly set (forcing the `resolver.getPythonPath(targetPythonVersion)` branch instead of the `PIP.getPythonPath` fallback) — supporting the root-cause analysis above.

---
*[View as Artifact](https://claude.ai/code/artifact/a2574f20-ba9b-423a-9e10-caa850f74a71)*

fdelbrayelle:
## Summary

- `PythonEnvironmentManager.setup()` keyed interpreter resolution for the Process runner only on whether `pythonVersion` was explicitly set, so declaring `dependencies` without `pythonVersion` still fell back to a hardcoded local PATH probe (`PackageManagerType.PIP.getPythonPath`), even when the dependencies were just installed against a different (e.g. `uv`-managed) interpreter.
- On a worker whose only Python is `uv`-managed (no local `python`/`python3` binary), this caused the script execution to fail with `python: not found` (exit code 127) despite the dependencies having installed successfully.
- Fix: key the branch on whether dependencies were actually installed (`!requirements.isEmpty()`), not just on `pythonVersion` being set. When either is true, the interpreter is resolved through `resolver.getPythonPath(...)`, which dispatches through the package manager that actually installed the dependencies (idempotent for `uv`, unchanged behavior for `pip`). The local-PATH probe is now reserved for the case with no explicit version and no dependencies, preserving #394's intent.
- Added a regression test asserting that with `packageManager=UV`, dependencies declared, and no explicit `pythonVersion`, the resolved interpreter is an absolute uv-managed path (not the bare `python` literal, and not a relative PATH lookup like `python3`).

fixes: https://github.com/kestra-io/plugin-scripts/issues/397

## Test plan

- [x] `rtk test ./gradlew :plugin-script-python:test --tests "io.kestra.plugin.scripts.python.internals.*" --tests "io.kestra.plugin.scripts.python.CommandsTest"` passes
- [x] Existing regression test `shouldResolveARunnableInterpreterForProcessRunnerWithoutExplicitPythonVersion` (no dependencies) still passes unmodified, confirming the local-probe branch is preserved
- [x] New test `shouldResolveTheManagedInterpreterForProcessRunnerWithDependenciesAndNoExplicitPythonVersion` fails without the fix (routes through the local PATH probe) and passes with it

---
*[View as Artifact](https://claude.ai/code/artifact/e7c6e755-47b7-409b-bd6a-16215bf9a577)*

### DIFF
--- a/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/internals/PythonEnvironmentManager.java
+++ b/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/internals/PythonEnvironmentManager.java
@@ -107,11 +107,14 @@ public class PythonEnvironmentManager {
         if (taskRunner instanceof Process || RunnerType.PROCESS.equals(runnerType)) {
             // The Process runner executes commands directly on the worker host, which may only ship
             // 'python3' (Debian/Ubuntu, the standard Kestra worker image) and not the bare 'python'
-            // literal. When an explicit pythonVersion is set, resolve it through the configured package
-            // manager (may provision a managed Python via 'uv'). Otherwise, probe for an interpreter
-            // that is already installed (python3, then python) rather than forcing a managed-Python
-            // download just to run a script with no dependencies.
-            pythonInterpreter = pythonVersion != null
+            // literal. Key this on whether dependencies were actually installed, not just on whether
+            // pythonVersion was explicitly set: dependencies may have just been installed against a
+            // managed interpreter (e.g. uv-managed) that has no relation to whatever 'python'/'python3'
+            // is on the local PATH, so the script must run with that same interpreter. Only fall back to
+            // probing for an already-installed local interpreter when there is truly nothing else to go
+            // on (no explicit version and no dependencies), to avoid forcing a managed-Python download
+            // just to run a dependency-free script.
+            pythonInterpreter = (pythonVersion != null || !requirements.isEmpty())
                 ? resolver.getPythonPath(targetPythonVersion)
                 : PackageManagerType.PIP.getPythonPath(resolver, targetPythonVersion);
         }
--- a/plugin-script-python/src/test/java/io/kestra/plugin/scripts/python/internals/PythonEnvironmentManagerTest.java
+++ b/plugin-script-python/src/test/java/io/kestra/plugin/scripts/python/internals/PythonEnvironmentManagerTest.java
@@ -1,5 +1,6 @@
 package io.kestra.plugin.scripts.python.internals;
 
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
@@ -18,6 +19,7 @@ import static io.kestra.core.utils.TestsUtils.mockRunContext;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.startsWith;
 
 @KestraTest
 class PythonEnvironmentManagerTest {
@@ -41,6 +43,37 @@ class PythonEnvironmentManagerTest {
         PythonEnvironmentManager.ResolvedPythonEnvironment environment = manager.setup(task.getContainerImage(), null, RunnerType.PROCESS);
 
         assertThat(environment.interpreter(), is(not("python")));
+
+        Process process = new ProcessBuilder(environment.interpreter(), "--version").start();
+        assertThat(process.waitFor(), is(0));
+    }
+
+    @Test
+    void shouldResolveTheManagedInterpreterForProcessRunnerWithDependenciesAndNoExplicitPythonVersion() throws Exception {
+        // Regression test for #397: with dependencies declared but no explicit pythonVersion, the
+        // interpreter used to be resolved via a hardcoded local PATH probe (PackageManagerType.PIP),
+        // regardless of the package manager that actually installed the dependencies. On a worker whose
+        // only Python is 'uv'-managed (no local python/python3 binary), that probe fails to find any
+        // interpreter at all, causing the script execution to fail ("python: not found", exit code 127).
+        // The interpreter must instead be resolved through the same package manager (UV here) that
+        // installed the dependencies, which returns the absolute path to the managed interpreter
+        // rather than a bare command name looked up on the local PATH.
+        Script task = Script.builder()
+            .id("python-env-manager-test-" + UUID.randomUUID())
+            .type(Script.class.getName())
+            .script(Property.ofValue("print('hello')"))
+            .packageManager(Property.ofValue(PackageManagerType.UV))
+            .dependencies(Property.ofValue(List.of("six")))
+            .build();
+        RunContext runContext = mockRunContext(runContextFactory, task, Map.of());
+
+        PythonEnvironmentManager manager = new PythonEnvironmentManager(runContext, task);
+        PythonEnvironmentManager.ResolvedPythonEnvironment environment = manager.setup(task.getContainerImage(), null, RunnerType.PROCESS);
+
+        assertThat(environment.interpreter(), is(not("python")));
+        // A bare PATH-looked-up command name (e.g. "python3") is relative; the UV-managed interpreter is
+        // always resolved to an absolute path, proving the resolution went through the package manager.
+        assertThat(environment.interpreter(), startsWith("/"));
 
         Process process = new ProcessBuilder(environment.interpreter(), "--version").start();
         assertThat(process.waitFor(), is(0));


fdelbrayelle:
**Edition**: EE | **Docker tag**: develop (next: 2.0.0-SNAPSHOT)
**Instance**: `s6x6x` — container `kestra-s6x6x`, reached directly at `http://localhost:8080` (the portless alias proxy on this shared dev box was cross-routing between concurrently running QA instances during this session, so direct host-port access was used instead; instance left running for manual review — `docker rm -f kestra-s6x6x` to stop).

## QA results summary

| Flow | Scenario | Result |
|------|----------|--------|
| `python_process_uv_deps_repro` | Process runner, `dependencies: [requests]`, `pythonVersion` unset, `packageManager` default (UV) | ✅ SUCCESS |
| `python_process_no_deps_no_version` | Process runner, no dependencies, `pythonVersion` unset (non-regression, #394) | ⚠️ FAILED — expected/unchanged behavior, not a regression (see notes) |

## Per-flow details

### Flow 1: python_process_uv_deps_repro (✅ SUCCESS)

<details>
<summary>Flow YAML</summary>

```yaml
id: python_process_uv_deps_repro
namespace: qa.issue397

tasks:
  - id: exec
    taskRunner:
    dependencies:
      - requests
    script: |
      import sys
      import requests
      print("python executable:", sys.executable)
      print("requests version:", requests.__version__)
```
</details>

**Gantt**
| Task | Status | Duration |
|------|--------|----------|
| `exec` | SUCCESS | 1.43s |
| **Total** | **SUCCESS** | **1.43s** |

**Logs synthesis**
```
WARN  Could not determine Python version automatically. Using fallback version '3.13' for dependency resolution and cache key computation. Set 'pythonVersion' explicitly if needed.
INFO  python executable: /tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13
INFO  requests version: 2.34.2
```
This is the exact reproducer scenario from the issue: `dependencies` declared, `pythonVersion` unset, worker has no local `python`/`python3` (only `uv`-managed). Before the fix this failed with `python: not found` (exit 127). After the fix, the script executes using the **uv-managed interpreter path** (`/tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13`) — the same interpreter `uv` provisioned to install `requests` — confirming the fix resolves the regression exactly as described in the plan.

**Outputs synthesis**
No declared outputs for this flow; stdout confirms correct interpreter and successful `requests` import.

### Flow 2: python_process_no_deps_no_version (⚠️ FAILED — expected, unchanged behavior)

<details>
<summary>Flow YAML</summary>

```yaml
id: python_process_no_deps_no_version
namespace: qa.issue397

tasks:
  - id: exec
    taskRunner:
    script: |
      import sys
      print("python executable:", sys.executable)
      print("no-dependency, no pythonVersion case still works")
```
</details>

**Gantt**
| Task | Status | Duration |
|------|--------|----------|
| `exec` | FAILED | 0.83s |
| **Total** | **FAILED** | **0.83s** |

**Logs synthesis**
```
WARN  Could not determine Python version automatically. Using fallback version '3.13' for dependency resolution and cache key computation.
ERROR Could not find a suitable Python interpreter for version: 3.13
```
This container image has **no local `python`/`python3` binary at all** (only `uv`). Per the plan's documented edge case ("No dependencies, no local Python, `packageManager=UV`: unchanged from today — still uses the local-probe fallback rather than forcing a managed-Python download"), this case intentionally still relies on the local-PATH probe and does **not** force a managed-Python download. On this particular container, that probe legitimately finds nothing, so the task fails — this is **identical, unchanged behavior from before the fix** (confirmed by the PR's own regression test `shouldResolveARunnableInterpreterForProcessRunnerWithoutExplicitPythonVersion`, which passed in CI against a host that does have local Python). Not a regression introduced by this PR.

## Timeout notes

None — both executions completed well within the 30-second timeout.

## Conclusion

The core fix is verified: with dependencies declared and no explicit `pythonVersion`, the Process runner now correctly executes using the `uv`-managed interpreter path instead of the bare `python` literal, resolving the `python: not found` (exit 127) regression from the issue. The no-dependency/no-version path is unchanged, as intended by the plan.

fdelbrayelle (commented):
## Kestra Plugin Code Review

*(Posted as COMMENT — GitHub does not allow self-approval since this PR is authored by the same account used for this review; verdict below is APPROVE.)*

Implementation matches the approved plan exactly: the interpreter-resolution ternary in `PythonEnvironmentManager.setup()` is now keyed on `(pythonVersion != null || !requirements.isEmpty())` instead of `pythonVersion != null` alone, so a task with dependencies declared but no explicit `pythonVersion` now resolves the interpreter via `resolver.getPythonPath()` (dispatching through the actually-configured package manager, e.g. UV's managed interpreter) instead of the hardcoded local-PATH `PackageManagerType.PIP` probe that caused exit 127 on uv-only hosts. All four edge cases called out in the plan/issue check out:
- `packageManager=PIP` explicit + deps + no version: routes to `resolver.getPythonPath()` → `PIP.getPythonPath()`, identical to the previous hardcoded call — no behavior change, confirmed by reading `PackageManagerType.PIP.getPythonPath`.
- Cache-hit path: `requirements` is rendered once at the top of `setup()` from the user-declared dependency list and is unaffected by whether `getPythonLibs` takes the cache-restore branch or the fresh-install branch, so `!requirements.isEmpty()` is evaluated identically in both cases — the new condition covers cache hits correctly.
- No deps + no version + UV configured: falls through to the local-probe `PackageManagerType.PIP.getPythonPath` branch, preserving #394's original intent (no forced managed-Python download for a dependency-free script) — matches the plan explicitly.
- Explicit `pythonVersion` set: unchanged, still routes through `resolver.getPythonPath()`.

### Kestra Guidelines — 1 finding
🟢 **Nit** — The new/updated inline comment block (5–6 lines) in `PythonEnvironmentManager.java` exceeds the 'one short line max' guidance for comments. Acceptable here since it documents a genuinely non-obvious invariant (a prior regression's root cause) that would otherwise likely be reintroduced, but consider trimming to the essential WHY if the file accumulates more such blocks.

### Security (OWASP Top 10:2025 + KPS) — no issues found
No new external input, string interpolation, process invocation, or secret-handling surface introduced by this diff — it is a pure boolean-condition change plus a test. No injection, path traversal, or credential-exposure risk.

### Performance — no issues found
No new allocation, loop, or blocking I/O introduced. The added branch is O(1) and reuses existing resolver calls.

### Test Quality
`shouldResolveTheManagedInterpreterForProcessRunnerWithDependenciesAndNoExplicitPythonVersion` correctly targets the regression: with `packageManager=UV`, `dependencies=["six"]`, and no `pythonVersion`, it asserts the resolved interpreter is neither the bare `"python"` literal nor a relative PATH lookup (`startsWith("/")`), then actually executes `<interpreter> --version` to confirm it's runnable. Under the pre-fix code this would have resolved via the hardcoded `PIP.getPythonPath` local probe, returning a relative name like `"python3"` (found on the CI host) rather than an absolute uv-managed path — so the `startsWith("/")` assertion would have failed pre-fix, confirming it is a genuine regression test. The existing no-dependency test is left untouched, correctly continuing to exercise the local-probe branch. Test pattern (installing a real lightweight dependency via uv) is consistent with existing precedent in `CommandsTest`/`ScriptTest`.

### Scope
Diff touches exactly the two files named in the issue/plan (`PythonEnvironmentManager.java` + its test) — no unrelated changes, no touched build files, no new dependencies.

**Verdict:** APPROVE

jymaire (approved):

PR #406, opened 2026-08-12
fix(python): honor an activated venv on the Process runner, with resolved-interpreter fallback

### LINKED ISSUE #404: Python Script: absolute uv-resolved interpreter path bypasses a venv activated in beforeCommands
## Summary

Since #397, `io.kestra.plugin.scripts.python.Script` on the **Process** task runner executes the script via the absolute uv-resolved interpreter path whenever `pythonVersion` is set. That bypasses any virtualenv activated in `beforeCommands`, so every import of a venv-installed package fails. `io.kestra.plugin.scripts.python.Commands` with byte-identical `beforeCommands` still works, because it passes user commands through verbatim.

This breaks the venv pattern documented at https://kestra.io/docs/how-to-guides/python-dependencies:

```yaml
beforeCommands:
  - python3 -m venv .venv
  - . .venv/bin/activate
  - pip install pandas kestra
```

Reported after an upgrade from Kestra 1.3.24 to 1.3.31.

## Environment

- **Kestra version:** 1.3.31 (EE) — worked on 1.3.24
- **Deployment:** Kubernetes, Postgres
- **Plugin:** plugin-scripts (bundled), regression present since ~v1.9.5 / #397
- **Task runner:** `io.kestra.plugin.core.runner.Process`
- **uv:** 0.6.17

## Steps to Reproduce

1. Set plugin defaults for `io.kestra.plugin.scripts.python` with the Process runner, an explicit `pythonVersion`, and a venv built in `beforeCommands`:

```yaml
- type: io.kestra.plugin.scripts.python
  forced: false
  values:
    taskRunner:
    pythonVersion: 3.13
    beforeCommands:
      - uv venv --python=3.13
      - uv pip install -r requirements.txt
      - . .venv/bin/activate
```

2. Store a `requirements.txt` namespace file containing a package that is absent from the worker host, e.g. `lxml>=6.0`.

3. Run a flow with one `Commands` task and one `Script` task that both import that package:

```yaml
id: tmp_py
namespace: company.team

tasks:
  - id: py_commands
    type: io.kestra.plugin.scripts.python.Commands
    logLevel: TRACE
    commands:
      - python -c 'import lxml'

  - id: py_script
    logLevel: TRACE
    script: |
      import lxml
```

## Expected Behavior

Both tasks resolve `lxml` from the venv created and activated in `beforeCommands`. A venv activated by the user should take precedence over the plugin's resolved interpreter path, or the plugin should at least make the venv's `site-packages` reachable.

## Actual Behavior

`py_commands` succeeds (exit 0). `py_script` fails with `ModuleNotFoundError: No module named 'lxml'`.

The generated shells differ only in the final line:

```
# py_commands — resolves through the activated venv's PATH
python -c 'import lxml'

# py_script — absolute path, so PATH (and therefore the venv) is bypassed
/tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13 /tmp/3Kd9d90JPg6p7FKvCA8CC1/10999592056222367262.py
```

## Root Cause Analysis

`PythonEnvironmentManager.setup()` (`plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/internals/PythonEnvironmentManager.java`, ~line 106):

```java
pythonInterpreter = (pythonVersion != null || !requirements.isEmpty())
    ? resolver.getPythonPath(targetPythonVersion)
    : PackageManagerType.PIP.getPythonPath(resolver, targetPythonVersion);
```

With `pythonVersion: 3.13` set, this takes the first branch, runs `uv python find 3.13 --system --python-preference=only-managed`, and returns an absolute path to the uv-managed interpreter. `Script.java` (~line 372) then builds the command as:

```java
.withCommands(Property.ofValue(List.of(
    String.join(" ", pythonEnvironment.interpreter(), commands.getTaskRunner().toAbsolutePath(...))
)))
```

`. .venv/bin/activate` only mutates `PATH` and `VIRTUAL_ENV` in the shell, so invoking an interpreter by absolute path never puts the venv's `site-packages` on `sys.path`. `Commands.java:349` uses `.withCommands(commands)` and is therefore unaffected.

#397 introduced the absolute path deliberately, to fix the case where the worker has no local Python and uv provisions a managed interpreter. The problem is the condition: `pythonVersion != null` also captures users who set `pythonVersion` but manage their own venv rather than using the `dependencies` property, which is a documented pattern.

Note that the plugin already has the mechanism needed to fix this — `Script.java` sets `PYTHONPATH` — but only when `pythonEnvironment.packages() != null`, i.e. only for `dependencies`-managed installs. A user-created venv gets no such treatment.

Possible directions:
- Detect an activated venv (`VIRTUAL_ENV` / `.venv/bin/python` in the working directory) and prefer its interpreter over the resolved one.
- Only force the absolute interpreter path when the plugin itself installed the dependencies (`!requirements.isEmpty()`), not merely because `pythonVersion` was supplied.
- Give users an explicit escape hatch to opt out of interpreter path resolution.

## Logs / Stack Trace

```
2026-08-11T12:00:43.181259Z DEBUG Finding Python '3.13' interpreter.
2026-08-11T12:00:43.191665Z DEBUG Executing command: uv python find 3.13 --system --python-preference=only-managed
2026-08-11T12:00:43.203912Z DEBUG /tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13
2026-08-11T12:00:43.406760Z DEBUG Starting command with pid 3630 [/bin/sh -c set -e
uv venv --python=3.13 2> /dev/null
uv pip install --index-url=<internal-mirror> -r requirements.txt 2> /dev/null
. .venv/bin/activate
find scripts/ -type f -name '*.py' -exec mv {} ./ \;
/tmp/python/cpython-3.13.3-linux-x86_64-gnu/bin/python3.13 /tmp/3Kd9d90JPg6p7FKvCA8CC1/10999592056222367262.py]
2026-08-11T12:00:43.811500Z ERROR Traceback (most recent call last):
2026-08-11T12:00:43.814620Z ERROR   File "/tmp/3Kd9d90JPg6p7FKvCA8CC1/10999592056222367262.py", line 4, in <module>
2026-08-11T12:00:43.814653Z ERROR     import lxml
2026-08-11T12:00:43.814659Z ERROR ModuleNotFoundError: No module named 'lxml'
2026-08-11T12:00:43.824565Z ERROR io.kestra.core.models.tasks.runners.TaskException: Command failed with exit code 1
	at io.kestra.plugin.scripts.exec.scripts.runners.CommandsWrapper.run(CommandsWrapper.java:216)
	at io.kestra.plugin.scripts.python.Script.run(Script.java:382)
	at io.kestra.plugin.scripts.python.Script.run(Script.java:34)
	at io.kestra.worker.WorkerTaskCallable.doCall(WorkerTaskCallable.java:85)
Caused by: io.kestra.core.models.tasks.runners.TaskException: Command failed with exit code 1
	at io.kestra.plugin.core.runner.Process.run(Process.java:152)
	at io.kestra.plugin.scripts.exec.scripts.runners.CommandsWrapper.run(CommandsWrapper.java:196)
```

## Workaround

Export `PYTHONPATH` to the venv's `site-packages` in `beforeCommands` after activation (`beforeCommands` and the interpreter invocation share one `/bin/sh -c` script). Safe only while the `dependencies` property is unused, since the plugin overwrites `PYTHONPATH` in that case:

```yaml
beforeCommands:
  - uv venv --python=3.13
  - uv pip install -r requirements.txt
  - . .venv/bin/activate
  - export PYTHONPATH="$(pwd)/.venv/lib/python3.13/site-packages"
```

## Customer Context

> Pylon ticket: #2170 | Priority: P2

Malaydewangan09:
## What

On the Process task runner, `io.kestra.plugin.scripts.python.Script` launched the script with the plugin's resolved interpreter by **absolute path**. When a venv was activated in `beforeCommands` (`. .venv/bin/activate`, which only mutates `PATH`/`VIRTUAL_ENV`), that absolute path bypassed the venv, so imports of venv-installed packages failed with `ModuleNotFoundError`. `Commands` was unaffected because it forwards user commands verbatim, so a bare `python` resolves through the activated `PATH`.

## Fix

`beforeCommands` and the run command execute in one shared shell, so the venv choice can be made at runtime. On the Process runner with no `dependencies`, prefer an activated venv, and fall back to the resolved interpreter otherwise:

```java
static String buildRunCommand(TargetOS os, TaskRunner<?> taskRunner, ResolvedPythonEnvironment pythonEnvironment, String scriptPath) {
    String resolvedInterpreter = pythonEnvironment.interpreter();

    // deps installed (packages resolved against this interpreter) or Windows -> unchanged
    if (pythonEnvironment.packages() != null || isWindowsTarget(os, taskRunner)) {
        return String.join(" ", resolvedInterpreter, scriptPath);
    }

    // no deps on POSIX -> use the venv if active, else the resolved interpreter
    return "if [ -n \"$VIRTUAL_ENV\" ]; then \"$VIRTUAL_ENV/bin/python\" \"" + scriptPath + "\"; else \"" + resolvedInterpreter + "\" \"" + scriptPath + "\"; fi";
}
```

The fallback is the **resolved** interpreter, which already probes `python3`/`python` and honors a pinned or managed `pythonVersion`. That keeps python3-only workers and pinned versions working when no venv is active. `PythonEnvironmentManager` interpreter resolution is unchanged.

## Behavior

| Case | Result |
|---|---|
| venv activated, no deps (the reported bug) | venv's python runs -> imports resolve |
| no venv, no deps, python3-only worker | resolved interpreter (`python3`) -> works |
| no venv, no deps, `pythonVersion` set | resolved managed interpreter -> pinned version honored |
| `dependencies` set | resolved interpreter + `PYTHONPATH`, unchanged |
| Windows | resolved interpreter, unchanged (no PowerShell branch) |

## Testing

- `ScriptTest.shouldUseActivatedVenvOverManagedInterpreterOnProcessRunner`: Process runner, `pythonVersion` set, no `dependencies`, a venv built and activated in `beforeCommands` installing a package, imported by the script. Asserts success.
- `buildRunCommand` unit tests: POSIX no-deps produces the `VIRTUAL_ENV` conditional with the **resolved interpreter** in the fallback branch; Windows and packages-present cases produce the single-token resolved-interpreter form.
- Full `:plugin-script-python:test`: all pass except a pre-existing `shouldExecScriptGivenDependency[DOCKER]` numpy/pandas cross-platform failure that also fails on `main` and runs on the unchanged `dependencies` path.

Fixes #404

### DIFF
--- a/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Script.java
+++ b/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Script.java
@@ -5,6 +5,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang3.SystemUtils;
+
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.models.annotations.Example;
 import io.kestra.core.models.annotations.Metric;
@@ -15,8 +17,10 @@ import io.kestra.core.models.executions.metrics.Counter;
 import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.RunnableTask;
 import io.kestra.core.models.tasks.runners.TargetOS;
+import io.kestra.core.models.tasks.runners.TaskRunner;
 import io.kestra.core.runners.FilesService;
 import io.kestra.core.runners.RunContext;
+import io.kestra.plugin.core.runner.Process;
 import io.kestra.plugin.scripts.exec.scripts.models.DockerOptions;
 import io.kestra.plugin.scripts.exec.scripts.models.ScriptOutput;
 import io.kestra.plugin.scripts.exec.scripts.runners.CommandsWrapper;
@@ -366,18 +370,15 @@ public class Script extends AbstractPythonExecScript implements RunnableTask<Scr
             env.put("PYTHONPATH", commands.getTaskRunner().toAbsolutePath(runContext, commands, relativePackagesPath.toString(), os));
         }
 
+        String scriptPath = commands.getTaskRunner().toAbsolutePath(runContext, commands, relativeScriptPath.toString(), os);
+        String runCommand = buildRunCommand(os, commands.getTaskRunner(), pythonEnvironment, scriptPath);
+
         ScriptOutput output = commands
             .addEnv(env)
             .withInterpreter(this.interpreter)
             .withBeforeCommands(beforeCommands)
             .withBeforeCommandsWithOptions(true)
-            .withCommands(
-                Property.ofValue(
-                    List.of(
-                        String.join(" ", pythonEnvironment.interpreter(), commands.getTaskRunner().toAbsolutePath(runContext, commands, relativeScriptPath.toString(), os))
-                    )
-                )
-            )
+            .withCommands(Property.ofValue(List.of(runCommand)))
             .withTargetOS(os)
             .run();
 
@@ -386,5 +387,37 @@ public class Script extends AbstractPythonExecScript implements RunnableTask<Scr
             pythonEnvironmentManager.uploadCache(runContext, pythonEnvironment.packages());
         }
         return output;
+    }
+
+    /**
+     * Builds the shell command that invokes the script.
+     * <p>
+     * When the plugin installed dependencies itself ({@code packages() != null}), PYTHONPATH already
+     * points to them and the resolved interpreter is authoritative: kept as the historical single-token
+     * invocation, unchanged.
+     * <p>
+     * Otherwise, a venv activated via {@code beforeCommands} (e.g. {@code . .venv/bin/activate}) only
+     * mutates the shell's PATH/VIRTUAL_ENV — it never changes which interpreter Kestra invokes. Since
+     * beforeCommands and this command run in the same shell, prefer that activated venv's python at
+     * runtime, falling back to the resolved interpreter (managed or bare) when no venv is active.
+     * <p>
+     * Package-private so the branching can be asserted directly without executing a process.
+     */
+    static String buildRunCommand(TargetOS os, TaskRunner<?> taskRunner, ResolvedPythonEnvironment pythonEnvironment, String scriptPath) {
+        String resolvedInterpreter = pythonEnvironment.interpreter();
+
+        if (pythonEnvironment.packages() != null) {
+            return String.join(" ", resolvedInterpreter, scriptPath);
+        }
+
+        return isWindowsTarget(os, taskRunner)
+            ? "if ($env:VIRTUAL_ENV) { & \"$env:VIRTUAL_ENV\\Scripts\\python.exe\" \"" + scriptPath + "\" } else { & \"" + resolvedInterpreter + "\" \"" + scriptPath + "\" }"
+            : "if [ -n \"$VIRTUAL_ENV\" ]; then \"$VIRTUAL_ENV/bin/python\" \"" + scriptPath + "\"; else \"" + resolvedInterpreter + "\" \"" + scriptPath + "\"; fi";
+    }
+
+    // Mirrors CommandsWrapper#getExitOnErrorCommands: targetOS is authoritative, AUTO only resolves to
+    // Windows when the worker itself runs Windows and the script executes directly on it (Process runner).
+    private static boolean isWindowsTarget(TargetOS os, TaskRunner<?> taskRunner) {
+        return os == TargetOS.WINDOWS || (os == TargetOS.AUTO && SystemUtils.IS_OS_WINDOWS && taskRunner instanceof Process);
     }
 }
--- a/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Script.java
+++ b/plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/Script.java
@@ -5,8 +5,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.apache.commons.lang3.SystemUtils;
-
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.models.annotations.Example;
 import io.kestra.core.models.annotations.Metric;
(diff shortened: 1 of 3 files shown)


Malaydewangan09:
Thanks for the careful review, all valid. Pushed a revision addressing each point.

**🔴 pythonVersion + no deps + no venv (pinned version ignored)** and **🔴 python3-only worker (`python: not found`)** — both fixed by not hardcoding `python`. The no-deps path now prefers an activated venv, and **falls back to the resolved interpreter** (which already probes `python3`/`python` and honors a pinned/managed `pythonVersion`) when no venv is active:

```java
if (packages() != null || isWindowsTarget(os, taskRunner)) {
    return String.join(" ", resolvedInterpreter, scriptPath);
}
return "if [ -n \"$VIRTUAL_ENV\" ]; then \"$VIRTUAL_ENV/bin/python\" \"" + scriptPath + "\"; else \"" + resolvedInterpreter + "\" \"" + scriptPath + "\"; fi";
```

- **Windows:** kept on the resolved interpreter (no PowerShell branch). The POSIX venv-detection has no reliable, tested PowerShell equivalent here, and a `Process` + Windows + venv case was neither reported nor exercised. Leaving Windows on the resolved interpreter means zero Windows behavior change.

**🟠 Stale `@Schema` docs** — updated both the class description and the `script` property description to note the script runs with the resolved interpreter, or an activated virtualenv when one is set up in `beforeCommands`.

**🟡 Wasted work** — resolved: the resolved interpreter is now the fallback branch, so it is no longer discarded.

**🟢 `var` nit** — the two new locals now use `var`.

**Missing test** — added `buildRunCommandPrefersActivatedVenvButFallsBackToResolvedInterpreterOnPosix`, which asserts the fallback branch uses the resolved interpreter (not bare `python`), plus Windows and packages-present cases assert the single-token resolved-interpreter form. The `VIRTUAL_ENV`-activated integration test is retained.

Re: `b154d419` — correct, that commit had the fallback right; this restores it (minus the guessed PowerShell). Full `:plugin-script-python:test` is green except the pre-existing `shouldExecScriptGivenDependency[DOCKER]` numpy/pandas cross-platform failure, which also fails on `main` and runs on the unchanged `dependencies` path.

PR #414, opened 2026-09-02
fix(groovy): default the Docker runner to root for Commands tasks

### LINKED ISSUE #411: [Bash2.0] Docker runner: .Commands runs as the image user in a root-owned working directory, so outputFiles fails
### Describe the issue

On the Docker task runner, the working directory is created as `root:root` but `.Commands` tasks run as the image user. On any non-root image the script cannot write to its own working directory, so `outputFiles` fails.

`.Script` tasks on the same image run as uid 0 and work fine.

This affects the groovy plugin's default image `groovy:jdk21` (`USER groovy`, uid 1000).

### Visuals and Reproducibility

Image

Run the flow below.

Actual: `uid=1000(groovy)`, dir `drwxr-xr-x 2 root root`, `FileNotFoundException: out.txt (Permission denied)`, `Captured 0 output file(s)`.
Expected: the file is written and captured.

The same write as a `groovy.Script` task on the same image logs `UID=0` and captures 1 file. Adding `user: "0"` to the runner also makes this pass.

```yaml
id: docker_workdir_user_mismatch
namespace: company.team

tasks:
  - id: commands_on_nonroot_image
    type: io.kestra.plugin.scripts.groovy.Commands
    taskRunner:
      type: io.kestra.plugin.scripts.runner.docker.Docker
    containerImage: groovy:jdk21
    outputFiles:
      - out.txt
    commands:
      - "id"
      - "ls -ld ."
      - "groovy -e 'new File(\"out.txt\").text = \"hello\"'"
```

### Environment

Kestra Version: 2.0.0-rc10, commit 9900dcf
Plugin Version: io.kestra.plugin.scripts 1.9.8
Setup: Standalone with Postgres, Kestra EE

jymaire:
Closes #411

## Problem

On the Docker task runner, the working directory can end up owned by a user other than the container image's default user. On a non-root image such as `groovy:jdk21` (`USER groovy`, uid 1000), a `groovy.Commands` task cannot write into its own working directory, so `outputFiles` fails with `FileNotFoundException: out.txt (Permission denied)` and captures 0 files.

`groovy.Script` already forces `user("root")` on its Docker runner and is therefore unaffected — `Commands` had no equivalent, which is the inconsistency reported in the issue.

## Fix

Default the Docker runner's `user` to `root` when the user has not configured one explicitly, on **both** config paths:

- the modern `taskRunner:` path, in `Commands.run()`;
- the deprecated `docker:` property path, in `Commands.injectDefaults()` — necessary because `CommandsWrapper.getTaskRunner()` rebuilds the runner from `dockerOptions` on every call and discards any `withTaskRunner(...)` override.

Unlike `Script` (which replaces the task runner wholesale and drops any custom Docker options), this preserves every other option the user configured — volumes, memory, cpu, pull policy, credentials — and never overrides an explicitly set `user`.

## Scope

Checked the other modules with a non-root-looking default image; both `oven/bun` and `denoland/deno` create a non-root account but never `USER`-switch to it, so they still run as root (`docker run --rm oven/bun id` → `uid=0(root)`, same for `denoland/deno`). Groovy is the only affected module.

## Tests

`plugin-script-groovy/src/test/java/io/kestra/plugin/scripts/groovy/CommandsTest.java`:

- `outputFilesOnNonRootImage()` — modern `taskRunner:` path
- `outputFilesOnNonRootImageLegacyDockerProperty()` — deprecated `docker:` path

Both build a throwaway image `FROM groovy:jdk21` with a user at uid 5000, so the reproduction does not depend on the host uid coincidentally matching the image's default uid (with `groovy:jdk21`'s own uid 1000 the bug reproduces only on some hosts/engines). Verified TDD-style: without the fix both fail with `uid=5000` and `Permission denied`; with it both log `uid=0(root)` and capture `out.txt`.

`./gradlew :plugin-script-groovy:test` — BUILD SUCCESSFUL, 3 tests, 0 failures.

## Note for reviewers

Running containers as root by default is a real behavior change for anyone relying on the image's non-root user. It matches what `groovy.Script` has always done, and setting `taskRunner.user` explicitly opts out. Happy to instead scope it behind something more explicit if you'd prefer.

### DIFF
--- a/plugin-script-groovy/src/main/java/io/kestra/plugin/scripts/groovy/Commands.java
+++ b/plugin-script-groovy/src/main/java/io/kestra/plugin/scripts/groovy/Commands.java
@@ -9,10 +9,13 @@ import io.kestra.core.models.annotations.PluginProperty;
 import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.RunnableTask;
 import io.kestra.core.models.tasks.runners.TargetOS;
+import io.kestra.core.models.tasks.runners.TaskRunner;
 import io.kestra.core.runners.RunContext;
 import io.kestra.plugin.scripts.exec.AbstractExecScript;
 import io.kestra.plugin.scripts.exec.scripts.models.DockerOptions;
 import io.kestra.plugin.scripts.exec.scripts.models.ScriptOutput;
+import io.kestra.plugin.scripts.exec.scripts.runners.CommandsWrapper;
+import io.kestra.plugin.scripts.runner.docker.Docker;
 
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotNull;
@@ -94,6 +97,10 @@ public class Commands extends AbstractExecScript implements RunnableTask<ScriptO
         if (original.getImage() == null) {
             builder.image(runContext.render(this.getContainerImage()).as(String.class).orElse(null));
         }
+        // mirrors the taskRunner-based default below, for the deprecated `docker` property path.
+        if (original.getUser() == null) {
+            builder.user("root");
+        }
         return builder.build();
     }
 
@@ -101,12 +108,19 @@ public class Commands extends AbstractExecScript implements RunnableTask<ScriptO
     public ScriptOutput run(RunContext runContext) throws Exception {
         TargetOS os = runContext.render(this.targetOS).as(TargetOS.class).orElse(null);
 
-        return this.commands(runContext)
+        CommandsWrapper commandsWrapper = this.commands(runContext)
             .withInterpreter(this.interpreter)
             .withBeforeCommands(beforeCommands)
             .withBeforeCommandsWithOptions(true)
             .withCommands(commands)
-            .withTargetOS(os)
-            .run();
+            .withTargetOS(os);
+
+        // Docker can create the working directory owned by a user other than the non-root image user (e.g. `groovy` on `groovy:jdk21`), breaking outputFiles - see #411.
+        TaskRunner<?> taskRunner = commandsWrapper.getTaskRunner();
+        if (taskRunner instanceof Docker docker && docker.getUser() == null) {
+            commandsWrapper = commandsWrapper.withTaskRunner(docker.toBuilder().user("root").build());
+        }
+
+        return commandsWrapper.run();
     }
 }
--- a/plugin-script-groovy/src/main/java/io/kestra/plugin/scripts/groovy/Commands.java
+++ b/plugin-script-groovy/src/main/java/io/kestra/plugin/scripts/groovy/Commands.java
@@ -29,7 +29,10 @@ import lombok.experimental.SuperBuilder;
 @NoArgsConstructor
 @Schema(
     title = "Execute Groovy files and commands",
-    description = "Runs Groovy commands or files in the JVM and captures their output."
+    description = """
+        Runs Groovy commands or files in the JVM and captures their output.
+
+        On the Docker task runner, the container runs as `root` unless `taskRunner.user` is set explicitly, so the task can always write into its working directory (which `outputFiles` requires). Set `taskRunner.user` to keep the image's own default user instead."""
 )
 @Plugin(
     examples = {
@@ -97,7 +100,7 @@ public class Commands extends AbstractExecScript implements RunnableTask<ScriptO
         if (original.getImage() == null) {
             builder.image(runContext.render(this.getContainerImage()).as(String.class).orElse(null));
         }
-        // mirrors the taskRunner-based default below, for the deprecated `docker` property path.
+        // Same root default as run(), for the deprecated `docker` property path - see #411.
         if (original.getUser() == null) {
             builder.user("root");
         }
@@ -115,7 +118,7 @@ public class Commands extends AbstractExecScript implements RunnableTask<ScriptO
             .withCommands(commands)
             .withTargetOS(os);
 
-        // Docker can create the working directory owned by a user other than the non-root image user (e.g. `groovy` on `groovy:jdk21`), breaking outputFiles - see #411.
+        // A non-root image user may not own the working directory, breaking outputFiles - see #411.
         TaskRunner<?> taskRunner = commandsWrapper.getTaskRunner();
         if (taskRunner instanceof Docker docker && docker.getUser() == null) {
             commandsWrapper = commandsWrapper.withTaskRunner(docker.toBuilder().user("root").build());
--- a/plugin-script-groovy/build.gradle
+++ b/plugin-script-groovy/build.gradle
@@ -18,4 +18,5 @@ dependencies {
     api group: 'org.apache.groovy', name: 'groovy-jsr223', version: '5.1.1'
 
     testImplementation project(path: ':plugin-script', configuration: 'testOutput')
+    testImplementation group: 'com.github.docker-java', name: 'docker-java-api', version: '3.6.0'
 }
--- a/plugin-script-groovy/src/main/resources/doc/io.kestra.plugin.scripts.groovy.md
+++ b/plugin-script-groovy/src/main/resources/doc/io.kestra.plugin.scripts.groovy.md
@@ -1,11 +1,17 @@
 # How to use the Groovy plugin
 
-Execute Groovy code directly in the Kestra JVM — no container or task runner required.
+Execute Groovy code in the Kestra JVM, or run Groovy scripts and commands on a task runner.
 
 ## Tasks
 
-`Eval` runs inline Groovy code and is the primary task for general scripting. `Script` runs a Groovy script file. Both execute in-process on the Kestra worker with access to the full JVM classpath — no `containerImage` or `taskRunner` is needed.
+`Eval` runs inline Groovy code and is the primary task for general scripting. It executes in-process on the Kestra worker with access to the full JVM classpath — no `containerImage` or `taskRunner` is needed.
+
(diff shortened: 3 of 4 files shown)


plugin-script-groovy/src/main/java/io/kestra/plugin/scripts/groovy/Commands.java
```diff
@@ -94,19 +97,30 @@ protected DockerOptions injectDefaults(RunContext runContext, DockerOptions orig
...
             .withBeforeCommands(beforeCommands)
             .withBeforeCommandsWithOptions(true)
             .withCommands(commands)
-            .withTargetOS(os)
-            .run();
+            .withTargetOS(os);
+
+        // Docker can create the working directory owned by a user other than the non-root image user (e.g. `groovy` on `groovy:jdk21`), breaking outputFiles - see #411.
+        TaskRunner<?> taskRunner = commandsWrapper.getTaskRunner();
+        if (taskRunner instanceof Docker docker && docker.getUser() == null) {
```
fdelbrayelle, line 123: 🟠 [Security/A02] Root-by-default now applies to every Docker image used by Commands
Problem: `docker.getUser() == null` defaults to "root" unconditionally for any containerImage/taskRunner the user supplies, not just groovy:jdk21. `commands` is a fully flow-author-controlled property (arbitrary shell/groovy), so this silently increases the blast radius of every Commands task on Docker, with no @Schema/doc note anywhere the user can see it before hitting this behavior.
Fix: Document the new default (class-level @Schema or the plugin's io.kestra.plugin.scripts.groovy.md gotcha section) so users relying on a non-root default can opt out via `user`, or narrow the fix to chown/chmod the working directory to the container's runtime uid instead of forcing root for the whole command.

fdelbrayelle (commented):
## Kestra Plugin Code Review

Issue #411: Docker runner creates the working directory as root while `Commands` runs as the image's non-root user, breaking `outputFiles` (matches `Script`, which already forces root). The fix covers both the modern `taskRunner:` path (`run()`) and the deprecated `docker:` path (`injectDefaults()`), and includes two regression tests. Scope was verified against bun/deno (unaffected), which is good discipline.

However: the linked issue's reporter environment is Kestra `2.0.0-rc10`. Decompiling `script-2.0.0-SNAPSHOT.jar` vs the currently-pinned `script-1.3.19.jar` (this repo's `gradle.properties`) confirms `AbstractExecScript` no longer declares a `docker` field / `getDocker()` in 2.0.0-SNAPSHOT — only the vestigial `injectDefaults(DockerOptions)` overload remains. That means the legacy-path half of this fix (`injectDefaults()` override, `Commands.java` lines 94-105) and its dedicated test `outputFilesOnNonRootImageLegacyDockerProperty()` (which calls `.docker(DockerOptions...)`, `CommandsTest.java` line 125) will not compile once this plugin bumps `kestraVersion` toward the 2.0 line the issue was actually filed against. Worth confirming which core version this fix is meant to ship against before merge.

### Kestra Guidelines — 3 findings
- 🟠 `Commands.java` line ~120: Undocumented behavior change — Docker now defaults to `root` for every `containerImage`/`taskRunner` combination used by `Commands` (not just `groovy:jdk21`), with no `@Schema` or plugin-doc (`io.kestra.plugin.scripts.groovy.md`) update explaining the new default.
- 🟡 `CommandsTest.java` line 151 (`buildNonRootTestImage`): test infra overengineering — a new test-only dependency (`docker-java-api`) plus building/pushing/removing a throwaway custom image just to get a non-root uid, when `DockerOptions.user`/`Docker.user` is a plain numeric-uid-capable String — `.user("5000")` on the stock `groovy:jdk21` image would reproduce the same bug without any of this.
- 🟢 `CommandsTest.java` line ~75-79: the 5-line comment blocks above both new test methods exceed the "one short line max" comment guideline; trim to a single line + issue link.

### Security (OWASP Top 10:2025 + KPS) — 1 finding
- 🟠 A02:2025 Security Misconfiguration (`Commands.java` line ~120): forcing every Docker-based `Commands` execution to run as `root` by default broadens the blast radius of the fully flow-author-controlled `commands:` property (arbitrary shell/groovy commands) for any custom image the user supplies, not just the reported one. This is the same trade-off already made in `Script.java`, but silently extending it to `Commands.java` without documentation removes the user's ability to know they need to opt out via `user`. The PR description itself invites this feedback ("Happy to instead scope it behind something more explicit") — recommend either documenting the new default prominently, or narrowing the fix to actually chown/chmod the working directory to the container's runtime user instead of running the whole command as root.

### Performance — 1 finding
- 🟡 `CommandsTest.java` line 163-164: `buildImageCmd(...).start().awaitImageId()` has no timeout; a stalled Docker build/daemon would hang the test run indefinitely. Not a production code path, low severity, and avoidable entirely per the test-simplification suggestion above.

**Verdict:** REQUEST CHANGES

*(Note: this review was submitted via the API in a degraded form — a diagnostic single-comment call was accidentally submitted as a review before the full multi-comment review could be posted; this body was then updated in place to carry the complete findings since GitHub does not allow converting an already-submitted review's event/state after the fact. Please treat the verdict above, not the "Commented" badge, as authoritative.)*

jymaire:
Thanks for the thorough review — pushed b0693ea addressing most of it. Point-by-point:

### 🟠 Root-by-default undocumented — fixed

Agreed, this was the weakest part of the PR. Now documented in two places:

- `Commands`' class-level `@Schema` description, so it shows up in the task's generated docs;
- a new **Container user** section in `io.kestra.plugin.scripts.groovy.md` explaining why the default exists and that `taskRunner.user` opts out.

I kept the root default rather than switching to a `chown`/`chmod` of the working directory. The chown approach needs a privileged step to run *before* the user's commands to be effective, so it either injects a root `beforeCommands` entry (same privilege, less visible) or needs core-side support in the Docker runner — both are a bigger change than this bug fix, and neither matches what `Script` already does. If we want to move away from root-by-default, it should be a deliberate change across both tasks rather than a divergence introduced here.

While in that doc I also corrected an unrelated inaccuracy it asserted: it claimed `Eval` and `Script` "both execute in-process on the Kestra worker", but `Script` runs on a task runner (Docker by default) — see `Script.run()`, which builds a `CommandsWrapper` and forces `Docker.builder().user("root")`.

### 🟡 `awaitImageId()` has no timeout — fixed

Bounded at 5 minutes (`IMAGE_BUILD_TIMEOUT_MINUTES`).

### 🟢 Comment blocks too long — fixed

Trimmed the two blocks in `CommandsTest` and both in `Commands.java` to a single line each, with the issue link.

### 🟡 `.user("5000")` instead of building a throwaway image — respectfully, this doesn't work

The fix only defaults to root **when `user` is unset** (`docker.getUser() == null` / `original.getUser() == null`). Passing `.user("5000")` explicitly opts out of the defaulting, so such a test would fail *with* the fix applied — it can't act as a regression test for it. The bug specifically requires the *image's own default user* to be non-root while the task leaves `user` unset, which is why the test builds an image with `USER kestratest` at uid 5000. Keeping it as is.

The uid-5000 image (rather than stock `groovy:jdk21` at uid 1000) is also deliberate: with uid 1000 the bug only reproduces on hosts whose own uid isn't 1000, so the test would silently pass on some machines and in some CI configurations.

### 🟠 Forward-compat with the 2.0 line — confirmed, and narrower than stated

Verified by decompiling the cached jars rather than trusting either of our recollections:

```
javap -p io.kestra.plugin.scripts.exec.AbstractExecScript
```

- `script-1.3.19` (what `gradle.properties` pins): declares `protected DockerOptions docker`, `getDocker()`, and both `injectDefaults(...)` overloads.
- `script-2.0.0-rc12`: `docker` / `getDocker()` are **gone**; both `injectDefaults(...)` overloads still exist, and `injectDefaults(RunContext, DockerOptions)` is only reachable from the deprecated overload.

So the correct read is:

- the `injectDefaults` **override in `Commands.java` still compiles on 2.0** — it just becomes unreachable, exactly like the pre-existing identical override in `Script.java` (line 90), which is in the same position and isn't a new problem introduced here;
- the only actual compile break on a 2.0 bump is the **test**'s `.docker(DockerOptions...)` builder call.

This PR targets the pinned `kestraVersion=1.3.19`, where the legacy `docker:` path is live and the fix is required for it — `CommandsWrapper.getTaskRunner()` rebuilds the runner from `dockerOptions` on every call, so a `withTaskRunner(...)` override alone doesn't survive on that path. When the repo bumps to the 2.0 line, that half and its test get deleted along with the `docker` field, same as every other legacy-`docker` usage across these modules. Happy to be corrected if the intent is to ship this against 2.0 directly — in that case I'd drop the `injectDefaults` half and the legacy test from this PR.

plugin-script-groovy/src/main/java/io/kestra/plugin/scripts/groovy/Commands.java
jymaire: Documented in b0693ea: the root default is now stated in `Commands`' class-level `@Schema` description (so it lands in the generated task docs) and in a new **Container user** section of `io.kestra.plugin.scripts.groovy.md`, which explains why the working directory may not be writable by the image user and that `taskRunner.user` opts out.

I did not take the chown/chmod alternative: to be effective it has to run before the user's `commands`, so it either injects a root-privileged `beforeCommands` step (same privilege, less visible to the flow author) or needs support in the core Docker runner. Both are larger than this fix, and either would leave `Commands` diverging from `Script`, which has always forced root. Moving off root-by-default seems worth doing deliberately for both tasks rather than only here — happy to open a follow-up issue for that if you agree.

fdelbrayelle (approved):

PR #418, opened 2026-09-07
fix(r): de-flake ScriptTest and stop building lubridate from source

jymaire:
## What

Two commits, both on `plugin-script-r`'s `ScriptTest`:

1. **`fix(r)`** — the test no longer scrapes the `WORKERTASKLOG` queue for the dates printed by the R script. The script writes them to an output file, and the test asserts on that file's content read back from internal storage.
2. **`perf(r)`** — lubridate is installed from Debian's prebuilt `r-cran-lubridate` package instead of with `install.packages()`.

## Why

The nightly `main` run [#34081882896](https://github.com/kestra-io/plugin-scripts/actions/runs/34081882896/job/101618584465) failed with:

```
java.util.concurrent.TimeoutException: Await failed to terminate within PT30S.
  at io.kestra.core.utils.Await.until(Await.java:47)
  at io.kestra.plugin.scripts.r.ScriptTest.script(ScriptTest.java:74)
```

The R task itself was fine — exit code 0, and the container stream shows the expected output right before the timeout:

```
* DONE (lubridate)
[1] "2010-06-04"
[1] "2011-06-04"
[1] "2012-06-04"
```

What failed is the `Await` on the log queue. Log emission is explicitly asynchronous (`LogEntryEmitter.emits`: *"the log entry is not guaranteed to be emitted immediately"*), and this test flooded it: `install.packages("lubridate")` found no CRAN binary for R 4.6.1 in the `r-base` image, so it built `cpp11`, `generics`, `timechange` and `lubridate` from source — ~90s and a few thousand log lines. The script's own three lines landed behind that backlog, so on a loaded runner they could arrive after the 30s window. Timing confirms it: last stdout line at `04:09:13`, timeout at `04:09:44`.

This test had already been patched twice for the same flakiness (b5d9521, 07f643b), each time by extending the wait. So this PR removes both halves of the race instead of widening the window again:

- Asserting on the output file is synchronous, and checks the exact three values in order rather than their presence somewhere in the logs.
- The Debian package removes the source build, which was both the log flood and the reason this module took ~2.5 minutes.

The script still writes its dates to stdout as well as to the file, so the `stdOutLineCount`/`stdErrLineCount` assertions stay meaningful now that the before-commands are quiet.

## Test

`./gradlew :plugin-script-r:test --tests '*r.ScriptTest*'` passes locally, and drops from **4m31s to 13s**.

### DIFF
--- a/plugin-script-r/src/test/java/io/kestra/plugin/scripts/r/ScriptTest.java
+++ b/plugin-script-r/src/test/java/io/kestra/plugin/scripts/r/ScriptTest.java
@@ -33,17 +33,20 @@ class ScriptTest {
 
     @Test
     void script() throws Exception {
-        // The dates are written to an output file rather than asserted on the task run logs: log
-        // emission is asynchronous, and the `install.packages("lubridate")` before-command builds
-        // lubridate and its dependencies from source, flooding the log queue with thousands of
-        // lines. The script's own output lands behind that backlog and made this test flaky.
+        // lubridate is installed from Debian's prebuilt `r-cran-lubridate` package rather than with
+        // `install.packages()`: the `r-base` image has no CRAN binary for its R version, so
+        // install.packages() builds lubridate and its dependencies from source (~90s and thousands
+        // of log lines). The dates are also written to an output file rather than asserted on the
+        // task run logs, as log emission is asynchronous and the script's own output could land
+        // behind that backlog, which made this test flaky.
         Script rScript = Script.builder()
             .id("r-script-" + UUID.randomUUID())
             .type(Script.class.getName())
             .beforeCommands(
                 Property.ofValue(
                     List.of(
-                        "Rscript -e 'install.packages(\"lubridate\")'"
+                        "apt-get update -qq",
+                        "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq r-cran-lubridate"
                     )
                 )
             )
@@ -52,14 +55,13 @@ class ScriptTest {
                 Property.ofValue(
                     """
                         library(lubridate)
-                        writeLines(
-                            c(
-                                as.character(ymd("20100604")),
-                                as.character(mdy("06-04-2011")),
-                                as.character(dmy("04/06/2012"))
-                            ),
-                            "dates.txt"
-                        )"""
+                        dates <- c(
+                            as.character(ymd("20100604")),
+                            as.character(mdy("06-04-2011")),
+                            as.character(dmy("04/06/2012"))
+                        )
+                        writeLines(dates)
+                        writeLines(dates, "dates.txt")"""
                 )
             )
             .build();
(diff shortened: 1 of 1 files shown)


fdelbrayelle (approved):
## Kestra Plugin Code Review

Test-only PR; no linked issue. Stated goal (de-flake `r.ScriptTest`) is met: the async log-queue race is replaced by a synchronous output-file assertion, and the root cause of the flood (source-compiled lubridate) is removed by installing the prebuilt Debian package. History confirms this test was patched twice before (b5d9521, 07f643b) for the same class of flakiness without addressing the root cause — this PR removes the race instead of widening the timeout again, which is the right fix.

### Kestra Guidelines — 2 findings (both non-blocking)
- 🟡 Comment length: the new 5-line comment above the `Script.builder()` call explains genuinely non-obvious context (async log emission + why Debian package vs source build), which is a legitimate exception per the 'workaround for external constraint' carve-out — but the style guide caps this at 'one short line max' for anything that isn't documenting an algorithm. Consider trimming to 1-2 lines and moving the fuller rationale to the commit message/PR body (already present there).
- 🟢 Nit: `String dates = new String(...)` should be `var dates = new String(...)` per the 'use var for all local variables whenever legally possible' rule. Pre-existing sibling tests (shell, go) inline the expression instead of binding a variable, so this is the first spot in the diff where the explicit type is avoidable.

Positive notes: unused imports (LogEntry, QueueInterface, Await, Flux, Duration, CopyOnWriteArrayList) correctly stripped; task id still randomized (`UUID.randomUUID()`); `.lines().toList()` follows the 'prefer .toList()' convention; the `StorageInterface`/`TenantService.MAIN_TENANT` output-file-read pattern matches established precedent in plugin-script-shell and plugin-script-go tests.

### Security (OWASP Top 10:2025 + KPS) — no issues found
No secrets, no new HTTP/network calls, no untrusted input reaching a shell command (before-commands are static literals). `apt-get install` runs inside the task's default `Docker` `taskRunner` (r-base image), not on the CI host, so there's no host-pollution or privilege concern.

### Performance — no issues found
This is a net performance win (source build removed, ~4m31s → ~13s locally per the PR description). The unclosed `InputStream` returned by `storageInterface.get(...).readAllBytes()` is a minor test-scope leak, but it exactly mirrors the existing convention in plugin-script-shell/plugin-script-go tests, so it isn't a regression introduced by this diff.

**Verdict:** APPROVE

jymaire:
Thanks for the review — both findings addressed in 747412f:

- 🟡 **Comment length**: trimmed the 6-line comment to 3 lines covering just the two non-obvious points (no CRAN binary for this R version → source build; logs are emitted asynchronously). The fuller rationale stays in the PR description and commit messages.
- 🟢 **`var`**: `String dates` → `var dates`.

Re-ran `./gradlew :plugin-script-r:test --tests '*r.ScriptTest*'` — passes in 16.6s.

On the two notes you flagged as non-regressions, I agree and left them as-is: the unclosed `InputStream` from `storageInterface.get(...)` mirrors the existing pattern in the shell/go tests, so changing it here would make this file inconsistent with its siblings rather than fixing anything.

PR #419, opened 2026-09-07
fix(python): give the all_python sanity check a 5-minute timeout

jymaire:
## What

Adds `timeout = "PT5M"` to `RunnerTest.all_python`, matching the go, deno, php, bun and dotnet sanity checks.

## Why

CI went red on `main` in [run 34103015639](https://github.com/kestra-io/plugin-scripts/actions/runs/34103015639) with a harness timeout, not an assertion failure:

```
Failed to resolve parameter [Execution arg0] in method
  [void io.kestra.plugin.scripts.python.RunnerTest.all_python(Execution)]:
Execution 5XFdbRli9DPS3MRvWa35Mx is currently at the status RUNNING which is not the awaited one
  at io.kestra.core.runners.TestRunnerUtils.awaitExecution(TestRunnerUtils.java:270)
```

The test ran 1m 01s against the 60s `TestRunnerUtils.DEFAULT_MAX_WAIT_DURATION`. Every task up to the cutoff had `SUCCESS` — the flow was simply still working through task seven of nine (`python_script_outputs`) when the harness gave up.

`all_python.yaml` runs nine sequential tasks: four `pip install kestra` in a fresh container, two `ghcr.io/kestra-io/pydata:latest` pulls, and two process-runner tasks that bootstrap `uv` (`Unable to detect an installed version of 'uv'. Attempting to install...`). It has been sitting at 85–90% of the budget for days:

| Run | When | Duration |
| --- | --- | --- |
| 33835569605 | Sep 4 | 50.9s ✅ |
| 34081882896 | Sep 7 04:06 | 53.8s ✅ |
| 34103015639 | Sep 7 09:04 | >60s ❌ |

The runs in between show `:plugin-script-python:test FROM-CACHE`, which is why this looks like a sudden break — the task had not actually re-executed since Sep 4.

## Scope

Python only. It is the one module still on the default timeout that is anywhere near the limit — `perl` 29.8s, `groovy` 33.3s, `lua` 11.3s, `nashorn` 7.5s, `shell` 6.0s in the same run. (`all_go` also hit 1m 00s there but passed, because it already declares `PT5M`.)

### DIFF
--- a/plugin-script-python/src/test/java/io/kestra/plugin/scripts/python/RunnerTest.java
+++ b/plugin-script-python/src/test/java/io/kestra/plugin/scripts/python/RunnerTest.java
@@ -14,7 +14,7 @@ import static org.hamcrest.Matchers.is;
 @KestraTest(startRunner = true)
 class RunnerTest {
     @Test
-    @ExecuteFlow("sanity-checks/all_python.yaml")
+    @ExecuteFlow(value = "sanity-checks/all_python.yaml", timeout = "PT5M")
     void all_python(Execution execution) {
         assertThat(execution.getTaskRunList(), hasSize(9));
         assertThat(execution.getState().getCurrent(), is(State.Type.SUCCESS));


fdelbrayelle (approved):
## Review

One-line, test-only diff: bumps the `@ExecuteFlow` timeout for the Python sanity check from the 60s default to 5 minutes, matching the existing pattern already used in go/deno/php/bun/dotnet modules.

**Business requirements — met**
Fixes CI flakiness caused by `TestRunnerUtils.awaitExecution` timing out while the execution was still `RUNNING`, not an assertion failure. Consistent with observed wall-clock time (50-54s) sitting close to the 60s ceiling. Scope stays limited to the Python module.

**Kestra guidelines — 0 findings**
Sanity-check `RunnerTest` using `@ExecuteFlow` + `@KestraTest(startRunner = true)`, matching documented convention. No production code touched.

**Security (OWASP Top 10:2025 + KPS) — 0 findings**
Test-only change, no secrets, no external calls added, no new dependency.

**Performance — 0 findings**
No production runtime impact; only affects local/CI test wall-clock ceiling.

**Non-blocking suggestion**
This is a symptom fix (CI budget) rather than root cause (growing wall-clock cost from pip install / image pulls / uv bootstrap on every run). If the trend continues, this test could eventually blow through 5 minutes too. Worth a follow-up issue to consider caching/pre-baking the Python test image — out of scope for this PR.

**Verdict: APPROVE**

Issue #313, opened 2025-12-24
Introduce `ScriptTrigger` & `CommandsTrigger` for each sub plugin


fdelbrayelle:
### Context

We want **each submodule of `plugin-scripts`** to expose **two [triggers](https://kestra.io/docs/workflow-components/triggers)**:

- `ScriptTrigger`
- `CommandsTrigger`

The goal is to allow a flow to be triggered whenever:
- the underlying **script command** is executed (`ScriptTrigger`)
- all the **commands** from the Commands task is executed (`CommandsTrigger`)

In addition, each trigger must support an **exit condition** so the trigger can decide whether it emits an event depending on the execution result.

Example: for `plugin-script-shell` we should be able to configure:
- `exitCondition: "exit 1"` (or similar) to trigger only on failed execution.

---

## ✅ What to implement

For **each sub-plugin in `plugin-scripts`**, add:

### 1. `ScriptTrigger`

- Triggers when the sub-plugin script execution is performed.
- Must emit an event containing:
  - `command` executed
  - `arguments`
  - `env` (if available)
  - `exitCode`
  - `stdout` / `stderr` (if capture is enabled)
  - relevant contextual metadata (execution id, namespace, flow id, etc.)

### 2. `CommandsTrigger`

- Triggers when the commands task is executed (the multi-commands task of the sub-plugin).
- Must emit an event containing:
  - list of commands executed (or the matched one)
  - exit code(s)
  - stdout/stderr if available
  - metadata (execution id, namespace, flow id, etc.)

---

## 🔧 Trigger configuration requirements

Each trigger should accept at least:

- `commandPattern` (string / regex / glob)
- `argumentsPattern` (optional)
- `exitCondition` (string expression)

Where `exitCondition` can match on:
- exit code (`exitCode == 1`)
- output (`stdout contains "..."`)
- or a literal shell-style condition (ex: `"exit 1"` for shell plugin)

### Example YAML

```yaml
triggers:
  - id: on_shell_fail
    type: io.kestra.plugin.scripts.shell.triggers.ScriptTrigger
    commandPattern: ".*"
    exitCondition: "exit 1"

triggers:
  - id: on_any_commands
    type: io.kestra.plugin.scripts.shell.triggers.CommandsTrigger
    commandPattern: ".*"
    exitCondition: "exit 0"
```

Pradumnasaraf:
[...]

PR #433, opened 2026-09-16
feat(bun): introduce ScriptTrigger & CommandsTrigger for the Bun plugin

### LINKED ISSUE #422: Introduce ScriptTrigger & CommandsTrigger for the Bun plugin
## Context

`plugin-script-bun` is missing both. Existing implementations in `plugin-script-go`, `plugin-script-node`, `plugin-script-python`, `plugin-script-ruby`, and `plugin-script-shell` (same package as the task classes, no nested subpackage) can be used as a reference.

### 1. `ScriptTrigger`

- Triggers when `io.kestra.plugin.scripts.bun.Script` execution completes.

### 2. `CommandsTrigger`

- Triggers when `io.kestra.plugin.scripts.bun.Commands` execution completes.

## Example YAML

```yaml
triggers:
  - id: on_bun_fail
    type: io.kestra.plugin.scripts.bun.ScriptTrigger
    commandPattern: ".*"

triggers:
  - id: on_bun_commands
    type: io.kestra.plugin.scripts.bun.CommandsTrigger
    commandPattern: ".*"
```

InnoxCodes:
## Summary

Fixes #422, part of #313. `plugin-scripts` wants every submodule to expose a `ScriptTrigger` and a `CommandsTrigger`, so a flow can be triggered whenever the underlying script/commands execution completes, with an exit condition deciding whether the trigger fires. `plugin-script-bun` was missing both.

## Implementation

- Supports edge mode (default on) to emit only on a not-matching-to-matching transition, so the trigger doesn't fire repeatedly on every poll while the condition stays true. The previous result is stored in the namespace KV store under a length prefixed key built from the flow and trigger ids, and it is updated on every poll, so it survives the trigger being rebuilt between polls
- Compiles the condition regex once and keeps it in a small bounded static cache, since the trigger itself is rebuilt on every poll

`ScriptTrigger` wraps `io.kestra.plugin.scripts.bun.Script`; `CommandsTrigger` wraps `io.kestra.plugin.scripts.bun.Commands`. Both default `containerImage` to `oven/bun`, matching the existing `Script`/`Commands` tasks' own default. The module doc has a Triggers section and `AGENTS.md` lists both classes.

An image that cannot be pulled (or any other error thrown before the task returns) is logged as a warning and does not count as a run: nothing is emitted and the edge state is left as it was.

## Testing

Ran locally with Docker, 31 tests across these five classes, all passing:

- `EdgeStateTest` (8): edge state in the KV store, no Docker needed. Each poll runs on a JSON round-tripped copy of the trigger, and there are cases for `edge: false`, for state being scoped per flow and per trigger, and for the key format. Removing the `kvStore.put` makes 4 of them fail
- `ConditionPatternTest` (3): the regex cache returns the same compiled pattern and stays correct past its size limit
- `CommandsTriggerTest` (4): same for `CommandsTrigger`. Its edge test now runs the second poll on a serialized copy and also fails without the KV write
- `./gradlew :plugin-script-bun:spotlessCheck` passes


plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/ScriptTrigger.java
```diff
@@ -0,0 +1,275 @@
...
+            When true (default), emit only on a transition from not matching to matching. When false, emit on every poll that matches.
+            """
+    )
+    @Builder.Default
+    @PluginProperty(group = "advanced")
+    protected Property<Boolean> edge = Property.ofValue(true);
+
+    @Builder.Default
+    @Getter(AccessLevel.NONE)
+    private final AtomicBoolean lastMatched = new AtomicBoolean(false);
```
jymaire, line 153: **Edge-mode state does not persist across polls.** `lastMatched` is a transient instance field, but Kestra re-instantiates the trigger from the flow definition on every polling evaluation, so it resets to `false` each poll. With `edge=true` (the default) and a persistently-matching condition, `!lastMatched.getAndSet(matched) && matched` is `true` on every poll, firing an execution each interval instead of only on the not-matching→matching transition. The unit test passes only because it reuses a single trigger instance. Edge state should be persisted via the trigger/context mechanism, not an in-memory field. (Same defect in `CommandsTrigger`.)

---

**Confirmed by QA** (OSS `develop`). The example flow below, with `edge: true`, generated **~8 executions in ~80s** — one per `PT10S` poll — instead of a single execution:

```yaml
id: bun_script_trigger
namespace: company.team
triggers:
  - id: script_failure
    type: io.kestra.plugin.scripts.bun.ScriptTrigger
    interval: PT10S
    exitCondition: "exit 1"
    edge: true
    containerImage: oven/bun
    script: |
      process.exit(1);
tasks:
  - id: log
```

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/CommandsTrigger.java
```diff
@@ -0,0 +1,275 @@
...
+            If false, the trigger emits on every poll where the condition matches.
+            """
+    )
+    @Builder.Default
+    @PluginProperty(group = "advanced")
+    protected Property<Boolean> edge = Property.ofValue(true);
+
+    @Builder.Default
+    @Getter(AccessLevel.NONE)
+    private final AtomicBoolean lastMatched = new AtomicBoolean(false);
```
jymaire, line 153: **Edge-mode `lastMatched` instance field does not persist across scheduler polls** (same defect as `ScriptTrigger`). The trigger is reconstructed each poll, so `lastMatched` is always `false` at evaluation time and `edge=true` fires on every matching poll rather than once per transition. E.g. `commands: [exit 1]` with `exitCondition: exit 1` generates an execution every interval.

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/CommandsTrigger.java
```diff
@@ -0,0 +1,275 @@
...
+
+        try {
+            ScriptOutput taskOutput = task.run(TriggerRunContext.forEmbeddedTask(runContext, task));
+            Integer exitCode = safeExitCode(taskOutput);
+            Map<String, Object> vars = safeVars(taskOutput);
+
+            return new Output(Instant.now(), renderedCondition, exitCode, vars);
+        } catch (RunnableTaskException e) {
+            ExtractedFailure failure = extractFailure(e);
+            return new Output(Instant.now(), renderedCondition, failure.exitCode, null);
```
jymaire, line 199: **On failure, `vars` is set to `null`, so regex/substring conditions can never match error output.** The schema says the condition is matched against vars "and error logs when the task fails (TaskException)", but the `RunnableTaskException` path passes `vars=null` and `buildHaystack` returns `""` when vars is null. Any non-`exit N` condition (e.g. `exitCondition: ERROR.*timeout`) against a failing command's logs always returns `false`, so the trigger never fires on failure unless the condition is `exit N`. Error logs are never actually added to the haystack. (Same in `ScriptTrigger`.)

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/ScriptTrigger.java
```diff
@@ -0,0 +1,275 @@
...
+            return taskOutput.getVars();
+        } catch (Exception ignored) {
+            return null;
+        }
+    }
+
+    private record ExtractedFailure(Integer exitCode) {
+    }
+
+    private ExtractedFailure extractFailure(RunnableTaskException e) {
```
jymaire, line 250: **`exit N` silently fails to match when no `TaskException` is in the cause chain.** `extractFailure` walks `e.getCause()` for a `TaskException`; if the failure is wrapped differently or is an infra error (e.g. image pull failure) with no `TaskException`, `exitCode` stays `null` and `matchesCondition` returns `false` for `exit 1`. A genuinely failed run then does not fire a trigger that was configured specifically to catch failures. (Same in `CommandsTrigger`.)

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/ScriptTrigger.java
```diff
@@ -0,0 +1,275 @@
...
+            return out.getExitCode() != null && out.getExitCode() == expected;
+        }
+
+        String haystack = buildHaystack(out);
+        if (haystack.isEmpty() || cond.isEmpty()) {
+            return false;
+        }
+
+        try {
+            return Pattern.compile(cond).matcher(haystack).find();
```
jymaire, line 218: **Minor efficiency:** the user-supplied condition regex is recompiled with `Pattern.compile(cond)` on every poll. The `exit N` pattern is correctly cached in a static field; the general-condition regex could be compiled once (e.g. cached) rather than on each `evaluate()` on the scheduler hot path. (Same in `CommandsTrigger`.)

plugin-script-bun/src/test/java/io/kestra/plugin/scripts/bun/CommandsTriggerTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Map<String, Object> triggerVars = execution.get().getTrigger().getVariables();
+        assertThat("condition should be present", triggerVars.get("condition"), is("toto"));
+        assertThat("exitCode should be present", triggerVars.get("exitCode"), notNullValue());
+        assertThat("exitCode should be 0", triggerVars.get("exitCode"), is(0));
+        assertThat("timestamp should be present", triggerVars.get("timestamp"), notNullValue());
+        assertThat("vars should be present", triggerVars.get("vars"), notNullValue());
+    }
+
+    @Test
+    void commandsTrigger_edgeModeShouldSuppressSecondEmission() throws Exception {
```
jymaire, line 77: **This test reuses a single trigger instance and so masks the edge-mode defect.** It calls `evaluate` twice on the same object, so `lastMatched` persists and the second emission is suppressed. Because the real scheduler reconstructs the trigger each poll, this test would still pass even though edge suppression is broken in production (see the `lastMatched` comment). To catch the real behavior, the second evaluation should use a freshly-built trigger instance.

jymaire (commented):
Automated correctness review (6 findings).

InnoxCodes:
Acknowledging this review. Findings 1, 2 and 6 (lastMatched not surviving the scheduler's per-poll reconstruction, and the test masking it by reusing one instance) are the exact same systemic defect a human reviewer independently caught on the sibling #436 (PowerShell) PR - confirmed real there, not fixed in place since it's copied from the already-merged go/node/python/ruby/shell plugins and needs a real cross-poll state mechanism, not a narrow per-plugin patch. I documented the limitation explicitly (code comment + corrected schema description + a note on the test that only proves same-instance dedup) rather than leave it silently broken, and I'll bring the same documentation fix here plus to #434/#435 in a follow-up pass, together with whatever the tracked `AbstractScriptTrigger` fix ends up being.

Finding 5 (regex recompiled every poll) is a fair, easy catch I'll fold into that same follow-up.

Finding 4 (`exit N` not matching when no `TaskException` is in the cause chain, e.g. an infra failure like an image-pull error) is a real gap distinct from the edge-mode issue - noted, needs its own look at what `extractFailure` should do when the cause chain doesn't contain a `TaskException` at all.

Finding 3 (vars=null on failure, so non-`exit N` conditions can never match a failing run) - same root cause as the "failure logs" doc overclaim on #436, which I corrected there by documenting the limitation rather than changing behavior. Will bring the same correction here.

🤖 AI-assisted - I read and independently verified this review's reasoning against the code rather than taking it at face value.

InnoxCodes:
Pushed 050d2fa, as promised in my earlier note. Going through the six findings:

1. **edge state does not persist across polls**: documented, not fixed. Code comment on `lastMatched`, corrected `edge` schema description in both triggers ("intended to", and that a distributed deployment still emits on every matching poll). A real fix needs a cross-poll state mechanism shared by the whole plugin family, which the maintainer is tracking separately.
2. **same defect in `CommandsTrigger`**: same treatment.
3. **`vars` is null on failure, so only `exit N` can match a failed run**: fixed the docs to say exactly that in both triggers. Behaviour unchanged.
4. **`exit N` not matching when no `TaskException` is in the cause chain**: not changed. Whether a container that never started should count as `exit 1` is a semantic choice that is identical across all the copies, so I'd rather see it decided once for the family.
5. **regex recompiled every poll**: left as is. It is negligible next to a container run per poll, and a static cache keyed on the rendered condition would need a bound because templated conditions can vary. Happy to add a bounded cache if wanted.
6. **`CommandsTriggerTest` edge test reuses one instance**: added a comment saying it only proves same-instance dedup. `ScriptTriggerTest` also stopped asserting against a private copy of the logic: `matchesCondition` is package-private and the tests call it on a real trigger.

Also `@EqualsAndHashCode.Exclude` on `lastMatched` and example intervals changed to `PT60S`.

13/13 unit tests pass locally, and breaking the exit-code branch of `matchesCondition` makes 4 of them fail. No Docker here, so the container-backed tests are unrun on my side.

Update: findings 1, 2, 5 and 6 are fixed for real in dada7bf (edge state in the namespace KV store, cached regex, edge tests on a serialized copy), and I have since run the container-backed tests locally (31 across the five trigger test classes, all passing). Findings 3 and 4 are still documented, see the review threads.

jymaire:
Hello @InnoxCodes , it will be difficult for me to merge this PR if all the bugs found during review are only documented and not fixed.

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/ScriptTrigger.java
InnoxCodes: Fixed in dada7bf. The previous result now lives in the namespace KV store, under a key built from the flow id length, the flow id and the trigger id, and it is written on every poll. I could not run a full Kestra instance here, so I covered your QA scenario with tests instead: each poll runs on a JSON round-tripped copy of the trigger, so with a script that keeps exiting 1 the first poll fires, the next ones do not, and it fires again after a poll where the condition did not match. Removing the `kvStore.put` makes those tests fail.

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/CommandsTrigger.java
InnoxCodes: Same fix here, see the ScriptTrigger thread. Fixed in dada7bf.

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/CommandsTrigger.java
InnoxCodes: This one is still a documented limitation, not a behaviour change. The schema and module doc no longer claim error logs are matched and say only `exit N` can match a failed run. Matching log lines is not possible from here: on failure the task only hands back a `TaskException` with the exit code and the stdout/stderr line counts, not the lines themselves. What it does hold is the structured `::{"outputs":...}::` values collected before the failure, so I could pass those as `vars` on a failed run. I held back because that changes what `trigger.vars` contains for failures. Tell me if you want it and I will add it.

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/ScriptTrigger.java
InnoxCodes: I tried an image that cannot be pulled (on the .NET copy of this same code) to see what happens. docker-java throws a `NotFoundException` before the task returns anything, so it is not a `RunnableTaskException` and never reaches `extractFailure`. `evaluate` logs "Trigger evaluation failed" as a warning and returns empty, and with the new edge state that poll also leaves the stored state alone. So an infrastructure error does not count as a run, which seems right to me since the script never exited with 1. I did not find another way to get a `RunnableTaskException` without a `TaskException` in it. If you would rather have those count as a failed run, it is a small change in `runOnce`, tell me which exit code it should report. I added a line about this to the PR description.

plugin-script-bun/src/main/java/io/kestra/plugin/scripts/bun/ScriptTrigger.java
InnoxCodes: Done in dada7bf. The compiled pattern is kept in a static map bounded to 64 entries (least recently used goes first), because the condition can be templated and vary between polls. `ConditionPatternTest` checks that the same pattern instance comes back and that it stays correct when more conditions than the limit go through.

plugin-script-bun/src/test/java/io/kestra/plugin/scripts/bun/CommandsTriggerTest.java
InnoxCodes: Agreed. The edge test now runs the second poll on a copy that went through JSON serialization, like the worker does, and `EdgeStateTest` does the same for every poll. Both fail if the KV write is removed. Trigger ids are also unique per test now, since the test storage keeps KV state between runs.

InnoxCodes:
Thanks @jymaire, fair point. I pushed dada7bf, which fixes what can be fixed in this PR: the edge state is now kept in the namespace KV store (the bug QA reproduced), the condition regex is cached, and the edge tests run every poll on a serialized copy of the trigger. It is all in the PR description too.

Two things are still documented rather than changed, and I explained why on the threads: a failed run has no vars, so only `exit N` can match it, and an image that cannot be pulled is logged as a warning instead of counting as a run. I can change either one if you tell me the behaviour you want.

jymaire (approved):
Thanks for bug fixing, I just renamed your PR to follow conventional commits (as described in [doc](https://kestra.io/docs/plugin-developer-guide/contribution-guidelines#general-guidelines)

PR #415, opened 2026-09-02
fix(deno): default permissions so env/inputFiles/outputFiles work on Script

jymaire:
## Summary

- `deno.Script` built `deno run <file>` with no permission flags. Deno denies env, filesystem read and write by default, so `env`, `inputFiles` and `outputFiles` all failed at runtime with `NotCapable` errors — the three features every other script task supports were unusable.
- Added a `permissions` property on `deno.Script` defaulting to `--allow-env`, `--allow-read`, `--allow-write`, giving parity with the rest of the module by default. Users can widen it (e.g. `--allow-net`) or set it to an empty list to run under Deno's secure-by-default sandbox.
- `deno.Commands` is unaffected — users already control the full command there.
- Updated the task Javadoc/`@Schema`, added a runnable example, and updated the plugin how-to doc.

closes: https://github.com/kestra-io/plugin-scripts/issues/410

## Test plan

- [x] `rtk test ./gradlew :plugin-script-deno:test` passes (includes a new test reproducing the issue — first confirmed it failed with `NotCapable: Requires env access to "MY_VAR"` against the old code, then verified it passes with the fix; also exercises inputFiles/outputFiles)
- [x] `rtk err ./gradlew :plugin-script-deno:compileJava :plugin-script-deno:compileTestJava` — no errors
- [ ] `./gradlew shadowJar`

PR #434, opened 2026-09-16
Introduce ScriptTrigger & CommandsTrigger for the Deno plugin

### LINKED ISSUE #423: Introduce ScriptTrigger & CommandsTrigger for the Deno plugin
## Context

`plugin-script-deno` is missing both. Existing implementations in `plugin-script-go`, `plugin-script-node`, `plugin-script-python`, `plugin-script-ruby`, and `plugin-script-shell` (same package as the task classes, no nested subpackage) can be used as a reference.

### 1. `ScriptTrigger`

- Triggers when `io.kestra.plugin.scripts.deno.Script` execution completes.

### 2. `CommandsTrigger`

- Triggers when `io.kestra.plugin.scripts.deno.Commands` execution completes.

## Example YAML

```yaml
triggers:
  - id: on_deno_fail
    type: io.kestra.plugin.scripts.deno.ScriptTrigger
    commandPattern: ".*"

triggers:
  - id: on_deno_commands
    type: io.kestra.plugin.scripts.deno.CommandsTrigger
    commandPattern: ".*"
```

InnoxCodes:
## Summary

Fixes #423, part of #313. `plugin-scripts` wants every submodule to expose a `ScriptTrigger` and a `CommandsTrigger`, so a flow can be triggered whenever the underlying script/commands execution completes, with an exit condition deciding whether the trigger fires. `plugin-script-deno` was missing both.

## Implementation

Ports the pattern already established in `plugin-script-go` and `plugin-script-node` (checked both as reference implementations, per the issue). A polling trigger that:

- Matches the rendered `exitCondition` against either the exit code (`exit N`) or a regex/substring against the emitted vars and failure logs
- Supports edge mode (default on) to emit only on a not-matching-to-matching transition, so the trigger doesn't fire repeatedly on every poll while the condition stays true

`ScriptTrigger` wraps `io.kestra.plugin.scripts.deno.Script`; `CommandsTrigger` wraps `io.kestra.plugin.scripts.deno.Commands`. Both default `containerImage` to `denoland/deno`, matching the existing `Script`/`Commands` tasks' own default.

## Testing

- `ScriptTriggerTest` mirrors the pure condition-matching unit tests from `plugin-script-go`'s `ScriptTriggerTest` (no Docker required), now calling the real package-private `matchesCondition` and covering the `permissions` wiring, 15/15 pass
- `CommandsTriggerTest` mirrors `plugin-script-go`'s integration-style tests that call `trigger.evaluate(...)` directly against a real container run. I don't have Docker available in my sandbox, so I could not run this file locally; it follows the exact same structure already merged and passing in `plugin-script-go`/`plugin-script-node`, with `denoland/deno:latest` as the image and shell-level commands (`exit 1`, `echo '::{"outputs":...}::'`) that run the same way regardless of the container's runtime
- `./gradlew :plugin-script-deno:compileJava :plugin-script-deno:compileTestJava` succeeds
- `./gradlew :plugin-script-deno:spotlessCheck` fails, but only on `Commands.java`, which has a pre-existing import-order violation on `main` unrelated to this PR

🤖 AI-assisted: I used Claude Code to help draft and verify this fix; I reviewed the diff and cross-checked it against the existing Go/Node reference implementations before opening this PR.


Malaydewangan09:
Same review as #436, because this is the same file with the package and image name swapped. 23 lines differ from the go copy, all cosmetic.

Ran it locally: 17/17 pass including the Docker tests.

Findings, all of which apply to the other copies too:

- edge mode doesn't survive the worker dispatch, so it fires on every matching poll
- `vars` is null on the failure path, so only `exit N` can ever match a failed run
- `extractFailure` returns a null exit code when there's no `TaskException` in the cause chain, so infra failures don't fire. Confirmed on this module: bad image tag with `exitCondition: exit 1` and `edge: false`, the trigger did not fire
- the user regex is recompiled on every poll
- `CommandsTriggerTest`'s edge test reuses one instance and masks the defect
- `ScriptTriggerTest` doesn't touch production code at all, it re-implements `matchesCondition` as `exitConditionMatches` and asserts against the copy

#436 already has fixes for the test and the doc wording, worth porting those across rather than re-deriving them.

One thing specific to Deno. `Script` has a `permissions` property (added in #415) defaulting to `--allow-env,--allow-read,--allow-write`. The trigger doesn't expose it, so a trigger script can never be granted `--allow-net`. Polling an HTTP endpoint is the obvious use for a Deno polling trigger and it's unreachable. Same script run both ways:

```
WITH --allow-net                    -> SUCCESS exitCode=0
trigger defaults (no --allow-net)   -> FAILED  exit code 1
```

That one is worth fixing here rather than deferring, since it's small and specific to this module.

InnoxCodes:
Thanks for the local run and for pointing at the #436 fixes, ported those across (936ae92, rebased onto current main).

**Deno specific, `permissions`**: fixed here as you suggested. `ScriptTrigger` now exposes the Script task's `permissions` property with the same default (`--allow-env`, `--allow-read`, `--allow-write`) and passes it through, so a trigger script can be granted `--allow-net`. I added an example that polls an HTTP endpoint with it. To unit test the wiring without Docker I moved the task construction into a package-private `buildTask()`; there are two tests, one that the default matches `Script`'s own default so they can't drift, and one that a custom value reaches the task. I checked the second one fails if I drop the passthrough. `CommandsTrigger` needs nothing since `Commands` takes the raw command line, so `deno run --allow-net ...` already works there.

**Ported from #436**: `ScriptTriggerTest` now calls the real package-private `matchesCondition` (CsvSource table for exit codes), the "failure logs" claim is gone from the docs (only `exit N` can match a failed run), the `edge` limitation is documented in the code comment, schema descriptions and on the edge test, `@EqualsAndHashCode.Exclude` on `lastMatched` with the corrected comment, and example intervals are `PT60S`. I also corrected the spotlessCheck line in the description: it only fails on `Commands.java`, which is already unformatted on main.

**Not changed, deliberately**:
- `extractFailure` returning a null exit code when there is no `TaskException`: this is a behaviour choice rather than a bug fix (should a container that never started count as `exit 1`?), and it is identical in all the copies, so I'd rather you decide it once for the family than have four PRs diverge.
- Regex recompiled per poll: I left it. Compiling a small pattern once a minute is negligible next to a container run, and a static cache keyed on the rendered condition would need a bound because templated conditions can vary. Happy to add a bounded cache if you'd still like it.

Unit tests pass locally (15/15 in this module). I still don't have Docker, so the container-backed tests are unrun on my side, though nothing in this change touches their paths apart from the new `permissions` default, which matches what `Script` already used.

Malaydewangan09:
`936ae92` is good. 23/23 pass here including the container-backed ones.

Checked the `permissions` work rather than just reading it:

- dropped `.permissions(this.permissions)` from `buildTask()` and `buildTask_passesTriggerPermissionsThroughToTheScriptTask` failed, so the passthrough is genuinely covered. The drift test still passes under that mutation, which is correct, it guards the default value rather than the wiring.
- end to end through `evaluate()` against a real container:

```
with --allow-net    fired=true   vars={net=status-200}
default perms       fired=false
```

So the gap is actually closed, not just wired.

The ported fixes all landed: no "failure logs" left in either class, no unqualified "emit only on transitions", `PT60S` in all three examples, `Exclude` on both, `matchesCondition` package-private on both.

Agreed on both deferrals. The `extractFailure` one is a real behaviour question (a container that never started isn't obviously `exit 1`) and should be decided once for the family. And you're right about the regex, a per-poll compile is noise next to a container start, and a naive cache keyed on a templated condition would grow unbounded. Leave it.

Nothing blocking from me.

Malaydewangan09 (approved):

PR #435, opened 2026-09-16
Introduce ScriptTrigger & CommandsTrigger for the .NET plugin

### LINKED ISSUE #424: Introduce ScriptTrigger & CommandsTrigger for the .NET plugin
## Context

`plugin-script-dotnet` is missing both. Existing implementations in `plugin-script-go`, `plugin-script-node`, `plugin-script-python`, `plugin-script-ruby`, and `plugin-script-shell` (same package as the task classes, no nested subpackage) can be used as a reference.

### 1. `ScriptTrigger`

- Triggers when `io.kestra.plugin.scripts.dotnet.Script` execution completes.

### 2. `CommandsTrigger`

- Triggers when `io.kestra.plugin.scripts.dotnet.Commands` execution completes.

## Example YAML

```yaml
triggers:
  - id: on_dotnet_fail
    type: io.kestra.plugin.scripts.dotnet.ScriptTrigger
    commandPattern: ".*"

triggers:
  - id: on_dotnet_commands
    type: io.kestra.plugin.scripts.dotnet.CommandsTrigger
    commandPattern: ".*"
```

InnoxCodes:
## Summary

Fixes #424, part of #313. `plugin-scripts` wants every submodule to expose a `ScriptTrigger` and a `CommandsTrigger`, so a flow can be triggered whenever the underlying script/commands execution completes, with an exit condition deciding whether the trigger fires. `plugin-script-dotnet` was missing both.

## Implementation

- Supports edge mode (default on) to emit only on a not-matching-to-matching transition, so the trigger doesn't fire repeatedly on every poll while the condition stays true. The previous result is stored in the namespace KV store under a length prefixed key built from the flow and trigger ids, and it survives the trigger being rebuilt between polls. The write is skipped when a poll repeats the previous result, so a condition that stays true for hours does not rewrite the same value every interval

`ScriptTrigger` wraps `io.kestra.plugin.scripts.dotnet.Script`; `CommandsTrigger` wraps `io.kestra.plugin.scripts.dotnet.Commands`. Both default `containerImage` to `mcr.microsoft.com/dotnet/sdk:10.0`, matching the existing `Script`/`Commands` tasks' own default. The module doc has a Triggers section and `AGENTS.md` lists both classes.

## Testing

Ran locally with Docker, 28 tests across these four classes, all passing:

- `EdgeStateTest` (10): edge state in the KV store, no Docker needed. Each poll runs on a JSON round-tripped copy of the trigger, and there are cases for `edge: false`, for state being scoped per flow and per trigger, for the key format, and for the write being skipped on a repeated result (checked against the KV entry's version, which stays put across repeats and increments on a real transition). Removing either `kvStore.put` makes the matching tests fail
- `CommandsTriggerTest` (4): same for `CommandsTrigger`. Its edge test also fails without the KV write
- `./gradlew :plugin-script-dotnet:spotlessCheck` fails, but only on `Script.java`, which has a pre-existing import-order violation on `main` unrelated to this PR


InnoxCodes:
Heads up for whoever reviews this: it is the same code as the PowerShell (#436), Deno (#434) and Bun (#433) ports, and those reviews found issues that apply here unchanged, so I ported the fixes before review (4c23436, rebased onto current main).

- `ScriptTriggerTest` now calls the real package-private `matchesCondition` instead of asserting against a copy of it. 13/13 pass locally, and breaking the exit-code branch makes 4 of them fail.
- The `edge` mode limitation is documented (code comment, schema descriptions, note on the edge test): `lastMatched` does not survive the worker's serialize/deserialize round trip, so a distributed deployment fires on every matching poll. The real fix needs a cross-poll state mechanism shared by the whole plugin family, tracked separately by the maintainers.
- The "failure logs" claim is removed from the docs (only `exit N` can match a failed run), `@EqualsAndHashCode.Exclude` is on `lastMatched`, and example intervals are `PT60S`.
- I corrected the spotlessCheck line in the description: it fails only on `Script.java`, which is already unformatted on main.

Left as in the sibling PRs, and open to changing once decided for the family: `extractFailure` returning a null exit code when there is no `TaskException`, and the per-poll regex compile. No Docker on my side, so the container-backed tests are unrun here.

Update: the edge limitation described above is fixed in 3cd505f. The state now lives in the namespace KV store, and I have since run the container-backed tests locally (28 across the four trigger test classes, all passing).

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
```diff
@@ -0,0 +1,292 @@
...
+    // on a not-matching-to-matching transition. Excluded from equals/hashCode so this
+    // mutable field itself never affects equality (equals/hashCode also always fall
+    // through to Object's reference identity via AbstractTrigger and this project's
+    // lombok.equalsAndHashCode.callSuper=call, so two identically built triggers are
+    // still unequal regardless - that part is a pre-existing, kestra-wide behavior,
+    // not something this exclusion changes).
+    @Builder.Default
+    @Getter(AccessLevel.NONE)
+    @EqualsAndHashCode.Exclude
+    private final AtomicBoolean lastMatched = new AtomicBoolean(false);
```
fdelbrayelle: 🔴 [Guidelines / Correctness] `edge` dedup does not survive worker dispatch — default behavior contradicts its own docs
Problem: `lastMatched` is a plain in-memory `AtomicBoolean` used to implement "edge mode" (emit only on a not-matching→matching transition), which is the property's **default** (`edge = true`). The field's own comment admits: "Polling triggers are dispatched to a worker as a serialized payload with no getter exposed for this field, so it never survives that round trip — in a real distributed deployment, edge mode degenerates to 'matched', firing on every poll." So in any real (non single-process-test) deployment, a flow with the default `edge: true` and a condition that stays true (e.g. a script that keeps failing) will fire a new execution on **every poll interval** instead of once on the transition — an execution storm, not the documented anti-spam behavior. This directly violates the mandatory guideline: "a trigger must never re-fire on an item it already delivered. Persist a watermark ... in the flow's namespace KV Store ... and advance it on every poll."
Fix: persist the match/no-match watermark via `runContext.namespaceKv(namespace)` (length-prefixed key built from `flowId`/`triggerId`) instead of an in-process `AtomicBoolean`, and advance it on every poll as required by the guideline. Note: this exact pattern is already merged in `plugin-script-go` and `plugin-script-node`'s `ScriptTrigger`/`CommandsTrigger` (pre-existing bug), so this PR propagates it to a third plugin rather than introducing it — but it should not be perpetuated further, and ideally the KV-backed fix lands here and is backported.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/CommandsTrigger.java
```diff
@@ -0,0 +1,289 @@
...
+    // on a not-matching-to-matching transition. Excluded from equals/hashCode so this
+    // mutable field itself never affects equality (equals/hashCode also always fall
+    // through to Object's reference identity via AbstractTrigger and this project's
+    // lombok.equalsAndHashCode.callSuper=call, so two identically built triggers are
+    // still unequal regardless - that part is a pre-existing, kestra-wide behavior,
+    // not something this exclusion changes).
+    @Builder.Default
+    @Getter(AccessLevel.NONE)
+    @EqualsAndHashCode.Exclude
+    private final AtomicBoolean lastMatched = new AtomicBoolean(false);
```
fdelbrayelle: 🔴 [Guidelines / Correctness] Same in-memory-only edge dedup as ScriptTrigger
Problem: identical `lastMatched` AtomicBoolean pattern — does not survive worker serialize/deserialize between polls, so `edge: true` (the default) does not actually dedupe in a real deployment; see the full explanation on the ScriptTrigger.java comment.
Fix: same fix — persist the watermark in namespace KV Store per the guideline instead of an in-memory field.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
fdelbrayelle: 🟠 [Guidelines] Missing `@ToString.Exclude` on mutable runtime-state field
Problem: `lastMatched` carries mutable runtime state (like `AtomicBoolean isKilled` in the guideline's own example) and has `@EqualsAndHashCode.Exclude` but no `@ToString.Exclude`. The class-level `@ToString` (mandatory on trigger-adjacent classes here) will therefore print this field's value whenever the trigger object is logged or dumped in an exception trace.
Fix: add `@ToString.Exclude` alongside the existing `@EqualsAndHashCode.Exclude` on this field, per "Exclude mutable lifecycle/state fields ... must be annotated `@EqualsAndHashCode.Exclude` and `@ToString.Exclude`."

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/CommandsTrigger.java
fdelbrayelle: 🟠 [Guidelines] Missing @ToString.Exclude on mutable runtime-state field
Problem: same as ScriptTrigger.java — `lastMatched` has `@EqualsAndHashCode.Exclude` but no `@ToString.Exclude`, so the class's Lombok-generated `toString()` prints this mutable field.
Fix: add `@ToString.Exclude` next to `@EqualsAndHashCode.Exclude` on this field.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
```diff
@@ -0,0 +1,292 @@
...
+import lombok.experimental.SuperBuilder;
+
+@SuperBuilder
+@ToString
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+@Schema(
+    title = "Trigger on .NET script condition",
+    description = "Polls by running an inline .NET script in a container (default image mcr.microsoft.com/dotnet/sdk:10.0) and emits when exitCondition matches. Edge mode is intended to emit only on transitions but currently does not survive a poll-to-poll worker dispatch (see the edge property). Polls every 60s by default. Accepts 'exit N' or a regex (fallback substring) matched against emitted vars; a failed run has no vars, so only 'exit N' can match a failure."
```
fdelbrayelle: 🟠 [Guidelines] Class-level `@Schema(description = ...)` uses a single-line string, not a text block
Problem: this ~500-character description is a plain `"..."` string. Guideline: "Use Java multiline string blocks (`"""`) for `@Schema(description = ...)`... Single-line strings are discouraged for documentation." Every field-level `@Schema` in this same file correctly uses `"""`, only the class-level one doesn't.
Fix: convert to a `"""..."""` text block, matching the field-level schemas below it.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/CommandsTrigger.java
```diff
@@ -0,0 +1,289 @@
...
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+@SuperBuilder
+@ToString
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+@Schema(
+    title = "Trigger a flow when .NET commands match a condition",
```
fdelbrayelle, line 47: 🟠 [Guidelines] Class-level @Schema(description = ...) uses a single-line string, not a text block
Problem: same as ScriptTrigger.java — the class-level description is a plain single-line string instead of a `"""` text block, inconsistent with every field-level @Schema in the same file.
Fix: convert to a `"""..."""` text block.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
```diff
@@ -0,0 +1,292 @@
...
+
+    private Output runOnce(RunContext runContext) throws Exception {
+        Script task = Script.builder()
+            .id(this.getId())
+            .type(Script.class.getName())
+            .containerImage(this.containerImage)
+            .script(this.script)
+            .build();
+
+        String renderedExitCondition = runContext.render(this.exitCondition).as(String.class).orElse("");
```
fdelbrayelle: 🟡 [Guidelines] `@NotNull` `exitCondition` rendered with `.orElse("")` instead of failing loudly
Problem: `exitCondition` is `@NotNull` (mandatory, no declared default), but a render-time empty result silently falls back to `""` rather than surfacing an actionable error. Per the guideline, `.orElse(default)` is for properties that declare a real default value; a mandatory property with no default should be validated/rendered with a named error instead of silently degrading to "never matches."
Fix: e.g. `.orElseThrow(() -> new IllegalArgumentException("exitCondition must be set"))`, or at minimum log a WARN when the rendered value is empty so a misconfigured flow doesn't just poll forever without ever firing. Non-blocking — mirrors the same pattern already merged in `plugin-script-go`/`plugin-script-node`.

plugin-script-dotnet/src/test/java/io/kestra/plugin/scripts/dotnet/ScriptTriggerTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.hamcrest.Matchers.is;
+
+/**
+ * Unit tests for ScriptTrigger's condition-matching logic.
+ *
+ * Calls the real, package-private ScriptTrigger#matchesCondition directly (same pattern as
+ * plugin-script-ruby's ScriptTriggerConditionTest), so a change to the production method is
+ * what these tests actually exercise, not a separately maintained copy of its logic.
+ */
+class ScriptTriggerTest {
```
fdelbrayelle, line 20: 🟠 [Guidelines] No happy-path test exercises `ScriptTrigger#evaluate()` end-to-end
Problem: this file only unit-tests the pure `matchesCondition(Output)` function; it never builds a `ScriptTrigger` and calls `trigger.evaluate(...)` against a real container run the way `CommandsTriggerTest` does for `CommandsTrigger`. So a bug in `runOnce()` — e.g. in `TriggerRunContext.forEmbeddedTask` wiring, `safeExitCode`/`safeVars`, or `extractFailure` — would not be caught by any test in this module. Per the guideline: "Every task and trigger must therefore have at least one unconditional happy-path test ... exercising its default code path."
Fix: add a Docker-backed test mirroring `CommandsTriggerTest#commandsTrigger_shouldTriggerOnImplicitFailureExit1` but calling `ScriptTrigger.builder()....build()` and `trigger.evaluate(...)`, asserting the execution fires with the expected `exitCode`/`condition` outputs.

fdelbrayelle (changes requested):
## Kestra Plugin Code Review

The PR delivers a `ScriptTrigger` and `CommandsTrigger` for `plugin-script-dotnet`, addressing the core ask of #424/#313 (poll the underlying Script/Commands task, emit an execution when a condition matches). However, neither issue's stated `commandPattern` / `argumentsPattern` properties nor the `exitCode != 0` expression-style `exitCondition` syntax shown in both issues' example YAML are implemented — only a literal `exit N` or regex/substring match against emitted vars is supported. This mirrors the precedent already merged in `plugin-script-go` and `plugin-script-node` (neither implements `commandPattern`/`argumentsPattern` either), so it's a pre-existing, repo-wide scope reduction rather than something unique to this PR — flagging for visibility, not blocking on it alone.

### Kestra Guidelines — 5 findings
- 🔴 Edge-mode dedup does not work as documented/defaulted (see inline, both trigger files)
- 🟠 Missing `@ToString.Exclude` on mutable `lastMatched` field (both trigger files)
- 🟠 Class-level `@Schema(description = ...)` uses a single-line string instead of a text block (both trigger files)
- 🟠 Package how-to doc not updated with a `## Triggers` section (cross-cutting, see below)
- 🟠 `AGENTS.md` not updated to list the new trigger classes (cross-cutting, see below)

**Cross-cutting (not anchorable to the new-file diff):**

1. 🟠 [Guidelines] Package how-to doc missing Triggers section
   Problem: `plugin-script-dotnet/src/main/resources/doc/io.kestra.plugin.scripts.dotnet.md` documents only `Script`/`Commands` under `## Tasks`; it has no `## Triggers` section for the two new trigger classes this PR adds.
   Fix: add a `## Triggers` section listing `ScriptTrigger`/`CommandsTrigger`, their key properties (`exitCondition`, `edge`, `interval`, `containerImage`) and outputs, per the "Plugin How-To Doc" guideline ("when a task, trigger... is added... update the matching section"). Note: `plugin-script-go`'s and `plugin-script-node`'s docs have the same gap — worth a follow-up across the repo, but this PR is the one adding new trigger surface right now.

2. 🟠 [Guidelines] AGENTS.md not updated for new trigger classes
   Problem: `AGENTS.md` §"Key Plugin Classes" → `plugin-script-dotnet` still lists only `Commands`/`Script`; the two new `ScriptTrigger`/`CommandsTrigger` classes aren't listed, so an agent reading `AGENTS.md` won't know they exist.
   Fix: add `io.kestra.plugin.scripts.dotnet.ScriptTrigger` and `io.kestra.plugin.scripts.dotnet.CommandsTrigger` to the `plugin-script-dotnet` entry (same gap exists for `go`/`node`, but keep this PR's own module in sync per the "Must Do" AGENTS.md rule).

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
No secret fields, no HTTP calls, no manual shell string concatenation (commands/script are handed to the existing `Commands`/`Script` tasks unchanged), no deserialization or XML parsing. Broad `catch (Exception e)` in `evaluate()` logs via `runContext.logger().warn(..., e)` without leaking any secret-shaped data.

### Performance — ✅ No issues found
Per-poll synchronous container execution matches the existing `go`/`node` pattern; the only dynamically-compiled `Pattern` is the user-supplied `exitCondition` regex, which is unavoidable per-evaluation compilation for a feature that accepts arbitrary user regex — the static exit-code pattern is correctly hoisted to a `static final`. No unbounded collections or leaked resources introduced.

**Verdict:** REQUEST CHANGES

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
InnoxCodes: Agreed, this was a real bug. The AtomicBoolean is gone and the previous result now lives in the namespace KV store, under a key built from the flow id length, the flow id and the trigger id (length prefixed so `a-b` + `c` and `a` + `b-c` cannot collide). It is written on every poll, including polls that do not match, so a condition that stays true fires once. If the container run itself errors, the state is left as it was.

I added `EdgeStateTest`, which runs each poll on a JSON round-tripped copy of the trigger (no Docker needed), and both edge tests that go through `evaluate` now do the same. All of them fail if I remove the `kvStore.put`. Fixed in 3cd505f.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/CommandsTrigger.java
InnoxCodes: Same fix here, see the reply on ScriptTrigger. Fixed in 3cd505f.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
InnoxCodes: The field is gone with the KV change, so there is nothing left to exclude.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/CommandsTrigger.java
InnoxCodes: Same as ScriptTrigger: the field is gone.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
InnoxCodes: Converted to a text block.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/CommandsTrigger.java
InnoxCodes: Converted to a text block.

plugin-script-dotnet/src/main/java/io/kestra/plugin/scripts/dotnet/ScriptTrigger.java
InnoxCodes: Now throws `IllegalArgumentException("exitCondition must render to a non-empty value")` when it renders empty or blank. It happens before the container starts, and `evaluate` already catches exceptions from `runOnce` and logs a warning, so a misconfigured flow shows the reason in its logs instead of polling forever without saying anything.

plugin-script-dotnet/src/test/java/io/kestra/plugin/scripts/dotnet/ScriptTriggerTest.java
InnoxCodes: Added `ScriptTriggerEvaluateTest` with three Docker-backed cases: `exit 1` emits with `condition` and `exitCode` 1, `exit 0` stays quiet, and edge mode suppresses the second poll on a round-tripped copy. I ran them locally and they pass.

fdelbrayelle:
@InnoxCodes Thanks for the fixes — most of #pullrequestreview-5266633325 is addressed (KV-store edge dedup, `@Schema` text blocks, `exitCondition` now `orElseThrow`, `## Triggers` doc section, AGENTS.md entries). Two remain open:

- 🟠 `lastMatched` still has only `@EqualsAndHashCode.Exclude`, missing `@ToString.Exclude` — in both `ScriptTrigger.java` and `CommandsTrigger.java`.
- 🟠 `ScriptTriggerTest` still only exercises `matchesCondition()`; no test calls `trigger.evaluate()` end-to-end the way `CommandsTriggerTest` does. Could you add a Docker-backed happy-path test for `ScriptTrigger#evaluate()`?

Can you take a pass at these two?

InnoxCodes:
Thanks for checking back. Both of these were actually already done in 3cd505f, the same commit you called the KV-store dedup fixed in:

`lastMatched` is not there anymore, the whole field is gone. There is nothing left to put `@ToString.Exclude` on. The dedup state lives in the namespace KV store now (`shouldEmit`/`edgeStateKey` in both trigger classes), which is presumably why you saw it as fixed.

`ScriptTriggerEvaluateTest.java` was added in that same commit and does exactly what you are asking for: three Docker-backed cases that call `trigger.evaluate(...)` end to end (fires on `exit 1`, stays quiet on `exit 0`, and suppresses a second poll in edge mode on a serialized copy of the trigger). I ran it locally along with the rest, 28 tests pass.

Both are still there on the current head (55d02e7, after your merge from main just now). Let me know if something did not come through on your side, happy to double check.

fdelbrayelle:
Verified against `dc1fc1b8` (current head). All 8 items from #pullrequestreview-5266633325 are closed:

- `lastMatched` removed, KV-store watermark confirmed correct across all four transition cases (false→false, false→true, true→true, true→false).
- `@ToString.Exclude` items: moot, confirmed — field no longer exists.
- Both class-level `@Schema` descriptions: confirmed text blocks.
- `exitCondition`: confirmed `orElseThrow`, with the WARN logged by `evaluate()`'s catch.
- `ScriptTriggerEvaluateTest`: confirmed genuinely end-to-end (not vacuous — traced that the two polls in the edge-suppression case hit the same KV key across the JSON round-trip).

One correction on your last comment: `dc1fc1b8` (one commit past what you described) changed the KV write from unconditional-every-poll to skip-if-repeated-result. Functionally still correct — no state loss, no TTL involved — but "written on every poll including non-matches" is now stale wording, worth fixing in the description if you touch this again.

One new (non-blocking) finding from the fix itself: `shouldEmit`, `edgeStateKey`, `matchesCondition`, `buildHaystack`, `safeExitCode`, `safeVars`, `extractFailure`/`ExtractedFailure` are byte-for-byte duplicated between `ScriptTrigger.java` and `CommandsTrigger.java`. That's new surface area from this round of fixes, not pre-existing. Both extend `AbstractTrigger` — could this move to a shared base class before merge? Not blocking, but the guidelines flag this exact pattern.

Otherwise LGTM from my side.

InnoxCodes:
Thanks for the thorough re-verification, and for catching the stale wording, fixed that inline comment.

Good catch on the duplication too, and you are right that it is new: those seven items were identical before too (same pattern already existed for the exit-code matching and the failure extraction), but the KV fix added `shouldEmit`/`edgeStateKey` as two more copies on top of that.

I held off on a shared base class in this PR for one reason: the exact same duplicated shape now exists identically in the sibling PowerShell PR (#436), since it got the same fix for the same review comment, and Bun (#433) had it before merging too. Pulling it into a shared base only in this PR would leave dotnet structurally different from its siblings for no functional reason, and a real fix should cover the whole family in one place, either an `AbstractScriptTrigger` in `plugin-script` or a small stateless helper class, not something I want to design unilaterally mid-review in a single-plugin PR. Ruby already has a TODO pointing at the same idea for a different reason (the family carries the exact same six methods copy-pasted six times over, not just two).

Happy to open a tracking issue for it if one does not already exist. Let me know if you would rather I take a pass at the extraction here instead.

fdelbrayelle (approved):

PR #436, opened 2026-09-16
Introduce ScriptTrigger & CommandsTrigger for the PowerShell plugin

### LINKED ISSUE #431: Introduce ScriptTrigger & CommandsTrigger for the PowerShell plugin
## Context

`plugin-script-powershell` is missing both. Existing implementations in `plugin-script-go`, `plugin-script-node`, `plugin-script-python`, `plugin-script-ruby`, and `plugin-script-shell` (same package as the task classes, no nested subpackage) can be used as a reference.

### 1. `ScriptTrigger`

- Triggers when `io.kestra.plugin.scripts.powershell.Script` execution completes.

### 2. `CommandsTrigger`

- Triggers when `io.kestra.plugin.scripts.powershell.Commands` execution completes.

## Example YAML

```yaml
triggers:
  - id: on_powershell_fail
    type: io.kestra.plugin.scripts.powershell.ScriptTrigger
    commandPattern: ".*"

triggers:
  - id: on_powershell_commands
    type: io.kestra.plugin.scripts.powershell.CommandsTrigger
    commandPattern: ".*"
```

InnoxCodes:
## Summary

Fixes #431, part of #313. `plugin-scripts` wants every submodule to expose a `ScriptTrigger` and a `CommandsTrigger`, so a flow can be triggered whenever the underlying script/commands execution completes, with an exit condition deciding whether the trigger fires. `plugin-script-powershell` was missing both.

## Implementation

- Supports edge mode (default on) to emit only on a not-matching-to-matching transition, so the trigger doesn't fire repeatedly on every poll while the condition stays true. The previous result is stored in the namespace KV store under a length prefixed key built from the flow and trigger ids, and it survives the trigger being rebuilt between polls. The write is skipped when a poll repeats the previous result, so a condition that stays true for hours does not rewrite the same value every interval

`ScriptTrigger` wraps `io.kestra.plugin.scripts.powershell.Script`; `CommandsTrigger` wraps `io.kestra.plugin.scripts.powershell.Commands`. Both default `containerImage` to `ghcr.io/kestra-io/powershell:latest`, matching the existing `Script`/`Commands` tasks' own default. The module doc has a Triggers section and `AGENTS.md` lists both classes.

## Testing

Ran locally with Docker, 28 tests across these four classes, all passing:

- `EdgeStateTest` (10): edge state in the KV store, no Docker needed. Each poll runs on a JSON round-tripped copy of the trigger, and there are cases for `edge: false`, for state being scoped per flow and per trigger, for the key format, and for the write being skipped on a repeated result (checked against the KV entry's version, which stays put across repeats and increments on a real transition). Removing either `kvStore.put` makes the matching tests fail
- `CommandsTriggerTest` (4): same for `CommandsTrigger`. Its edge test now runs the second poll on a serialized copy and also fails without the KV write
- `./gradlew :plugin-script-powershell:spotlessCheck` fails, but only on `Commands.java`, which has a pre-existing violation on `main` unrelated to this PR


Malaydewangan09:
Ran this locally. All 17 tests pass including the 4 Docker ones you couldn't run, the powershell image is fine and both test commands behave as expected under pwsh. So the thing works.

Two things I'd want fixed before merge.

**`edge` doesn't actually do anything.** Polling triggers get dispatched to a worker as a serialized payload, and `lastMatched` is an `AtomicBoolean` with no getter, so it never survives the hop. The serialized trigger:

```
{"id":"t","type":"...ScriptTrigger","script":"exit 1","exitCondition":"exit 1"}
```

No `lastMatched`. Simulating the per-poll round trip the worker does:

```
poll#1 edge=true fired=true
poll#2 edge=true fired=true
poll#3 edge=true fired=true
```

It re-fires on every poll, which is the opposite of what the property promises. And `commandsTrigger_edgeModeShouldSuppressSecondEmission` only passes because it reuses the same object in memory. It's green while asserting something that never holds in production. That bothers me more than the bug itself, because it'll block whoever tries to fix it later.

**`ScriptTriggerTest` doesn't test `ScriptTrigger`.** It re-implements `matchesCondition` as `exitConditionMatches` inside the test and asserts against the copy. The real method is private and never called, so you could empty out the body and all 9 stay green. Ruby already does this right in `ScriptTriggerConditionTest`: makes `matchesCondition` package-private and calls it directly with a `@CsvSource` table. Worth copying.

Smaller stuff:

- Docs say the condition is matched "against emitted vars and failure logs". There are no failure logs. `buildHaystack` only returns `vars.toString()`, and the failure path passes `null` vars, so on a failed run nothing except `exit N` can ever match. Inherited from the others, but let's not add a sixth copy of a wrong doc.
- `@EqualsAndHashCode` includes the `AtomicBoolean`, so two identically built triggers aren't equal and hash differently. Needs `@EqualsAndHashCode.Exclude`.
- The trigger only exposes `containerImage` and `script`/`commands`. No `env`, `inputFiles`, `namespaceFiles`, or docker options, so no secrets and no memory limits on a container that restarts every 60s forever. Also means `CommandsTrigger` can't run an existing `.ps1` from namespace files, which is the main use of the Commands task.
- You used go as the reference, which is the oldest of the five. Python (#338) already factored this into `AbstractPythonTrigger` and comes in at 83 lines instead of 275, and ruby's `ScriptTrigger` has a literal TODO saying to extract this into an `AbstractScriptTrigger` in `plugin-script`.
- `spotlessCheck` doesn't pass, but it's pre-existing on `Commands.java` in main, not your files. Just worth correcting in the description.
- Example uses `PT10S`. Cold start was 46s for me, warm runs 2-4s. `allowConcurrent` is false by default so nothing overlaps, but it reads optimistic.

The duplication is not on you, it's on us for letting it reach five copies. I'll open a separate issue for the `AbstractScriptTrigger` refactor so this PR doesn't have to carry it.

InnoxCodes:
Thanks for actually running this and for the depth here. Confirmed both main points independently before touching anything, pushed a fix for what's safely scoped to this PR:

**edge / lastMatched**: confirmed your finding exactly - it's a plain `AtomicBoolean` with no exposed getter, so it doesn't survive the worker's serialize/deserialize round trip, and edge mode degenerates to "fires on every matching poll" in a real distributed deployment. I didn't attempt a full fix here: this exact pattern is copied from the already-merged go/node plugins and exists identically in python/ruby/shell too, so a correct fix needs an actual mechanism for persisting state across a serialized worker dispatch, which is a design question for the whole family, not something to solve unilaterally in a PowerShell-port PR. What I did do:
- Added an explicit code comment on `lastMatched` stating the limitation directly (dispatched as a serialized payload, no getter, doesn't survive the round trip, degenerates to firing every matching poll).
- Corrected the `edge` property's schema description to say "intended to" rather than asserting it works.
- Added the same limitation note on `commandsTrigger_edgeModeShouldSuppressSecondEmission`, since it's still only proving same-instance dedup.

Happy to pick this up properly once you've opened the `AbstractScriptTrigger` tracking issue - would rather fix it once, correctly, across all five, with your input on the right persistence mechanism, than patch around it here.

**ScriptTriggerTest doesn't test ScriptTrigger**: fixed by copying Ruby's `ScriptTriggerConditionTest` pattern exactly - `matchesCondition` is now package-private, the test builds a real `ScriptTrigger` and calls it directly, and the exit-code cases are a `@CsvSource` table. Verified this actually tests production code now: temporarily broke `matchesCondition`'s exit-code branch, confirmed the test caught it, reverted.

**Smaller stuff**, all fixed:
- "failure logs" removed from both trigger's docs - `buildHaystack` only ever returns `vars.toString()`, and the failure path passes null vars, so only `exit N` can ever match a failed run. Documented that explicitly instead.
- `@EqualsAndHashCode.Exclude` added on `lastMatched` on both triggers.
- Both examples' `interval` changed from `PT10S` to `PT60S`, matching the coded default and your measured cold start.
- PR description spotlessCheck claim: you're right, it's `Commands.java` on `main`, unrelated to my files - I'll fix the description.

Not touched, intentionally deferred rather than silently dropped:
- `env`/`inputFiles`/`namespaceFiles`/docker options - agreed this is a real gap (no secrets, no memory limits, `CommandsTrigger` can't run an existing `.ps1`). This felt like scope creep for a trigger-parity PR and inherits the same gap from go/node, so I'd rather do it as a deliberate follow-up than rush it in here.
- Adopting the leaner `AbstractPythonTrigger`/Ruby TODO pattern - agreed, but that's exactly the refactor you said you'd track separately, so I left this PR matching the existing (if imperfect) go/node shape rather than defining a new abstraction unilaterally.

All 13 non-Docker `ScriptTriggerTest` tests pass locally against the real method; still don't have Docker available to re-run the 4 integration tests, but nothing in this diff touches their code paths.

🤖 AI-assisted, as disclosed in the PR description - I read and verified all of the above myself rather than taking your comment at face value.

Malaydewangan09:
Pulled `2194648`, re-ran everything. 21/21 including the Docker tests.

Verified fixed: `matchesCondition` is package-private and the test calls it for real (mutation-tested it, broke the exit-code branch, 4 tests failed). `PT10S` to `PT60S`. `Exclude` is on the field and the bytecode confirms `equals()` no longer touches `lastMatched`. Honest wording on `edge`.

One miss: `ScriptTrigger.java:47`, the class-level `@Schema(description = ...)`, is unchanged from the first commit. Still says "Supports edge mode to emit only on transitions ... matched against emitted vars and failure logs", both claims you corrected everywhere else. That's the description rendered in the UI, so it's the most-read one of the set. `CommandsTrigger`'s is clean.

Correction on my side: the `Exclude` works but doesn't do what its comment claims. Two identically built triggers are still unequal. Real cause is `lombok.equalsAndHashCode.callSuper = call` in our `lombok.config` plus `AbstractTrigger` having no `equals`, so every trigger's `equals` opens with an `Object.equals` identity check. Kestra-wide, not yours, and nothing depends on it. Keep the `Exclude`, just drop the "so two identically built triggers still compare equal" part of the comment.

Fix line 47 and I'm happy. The deferrals are the right calls, tracking issue coming.

InnoxCodes:
Thanks for the mutation test and for catching both of these.

**Class-level description**: you're right, `ScriptTrigger.java:47` was unchanged since the first commit - fixed now to match the corrected wording on `edge`/`exitCondition`. `CommandsTrigger`'s was already accurate, left as-is.

**The `Exclude` comment**: verified your explanation independently before touching anything - checked `lombok.config` (`equalsAndHashCode.callSuper = call`) and `AbstractTrigger` (no `equals()` override, so it's plain `Object` identity). You're right that `callSuper` means every generated `equals()` opens with that identity check regardless of what's excluded, so two "identically built" triggers are never equal no matter what I do here - that's pre-existing and kestra-wide, not something this PR touches or fixes. Kept `@EqualsAndHashCode.Exclude` (still correct: it keeps this specific mutable field out of the comparison, which is the right call for a field mutated during evaluate()), corrected the comment to describe only what it actually does instead of the broader claim.

🤖 AI-assisted - verified both independently against the actual lombok config and AbstractTrigger source rather than taking the correction at face value.

Malaydewangan09:
`38d9065` looks right. Class-level description is honest now and the `Exclude` comment describes what it actually does. Both verified.

Two things still open:

- `extractFailure` only picks up an exit code when a `TaskException` is in the cause chain. If the container never starts (bad image, pull failure) there isn't one, `exitCode` stays null, and `exit 1` doesn't match. A trigger configured specifically to catch failures stays silent on a real failure. I confirmed this on the Deno copy of the same code: bad image tag, `exitCondition: exit 1`, `edge: false`, trigger did not fire.
- The user-supplied regex is recompiled on every poll. `EXIT_CONDITION_PATTERN` is cached in a static, this one isn't.

Neither is introduced by you and both exist in all six copies, so I'm fine treating them as part of the same family cleanup rather than blocking here. Flagging so they don't get lost.

InnoxCodes:
@Malaydewangan09 heads up, I pushed d29797d after your last look. It moves the `edge` state out of the `AtomicBoolean` and into the namespace KV store (key built from the flow id length, flow id and trigger id, written on every poll), so the re-fire you reproduced on every poll no longer happens. The maintainer review on the .NET port asked for exactly this, so I applied it here too instead of waiting for the shared refactor.

Beyond that it adds `EdgeStateTest` (runs each poll on a serialized copy of the trigger, no Docker), a Docker backed `ScriptTrigger#evaluate` test, makes the edge test in `CommandsTriggerTest` use a serialized copy for the second poll, and has an empty `exitCondition` fail with a clear message. I ran everything locally with Docker: 28 tests across the four trigger test classes pass, and removing the KV write makes the edge tests fail. The `extractFailure` null exit code and the per-poll regex compile are still as they were, since you were fine leaving those for the family cleanup.

Malaydewangan09:
`d29797d` fixes it. 32/32 pass here with Docker.

I re-ran the reproduction that started this, full `evaluate()` path against a real container, trigger reserialized every poll, stable `TriggerContext`:

```
edge=true   poll1=FIRED  poll2=quiet  poll3=quiet
edge=false  poll1=FIRED  poll2=FIRED
```

That was `FIRED FIRED FIRED` before. The KV key construction is safe: flow ids are `^[a-zA-Z0-9][a-zA-Z0-9._-]*` and trigger ids `^[a-zA-Z0-9][a-zA-Z0-9_-]*`, both subsets of the KV key charset, and the length prefix does rule out the collision you mention. `EdgeStateTest` is a good test, the poll sequence covers the transitions that matter.

Three small things on the KV handling, none blocking:

- `shouldEmit` is called outside the try/catch that wraps `runOnce`. A container failure is swallowed and returns empty, but a KV read or write failure propagates out of `evaluate()`. That asymmetry looks unintended, though failing loudly on a KV outage is arguably the better half of it.
- `kvStore.put` runs on every poll even when the value is unchanged. A trigger on the default 60s interval writes ~1440 times a day to say the same thing. Writing only when `matched != previouslyMatched` would cut that to transitions.
- The entry has no TTL, so it outlives the flow it belongs to. Not harmful, just never cleaned up.

Everything else from the .NET review landed here too: class-level `@Schema` is a text block, `## Triggers` is in the package doc, `AGENTS.md` lists both classes, `exitCondition` throws instead of degrading to `""`, and `ScriptTriggerEvaluateTest` covers the happy path through `evaluate`.

Good from me.

InnoxCodes:
Thanks for the close re-check, and for calling out the reproduction numbers, that is exactly the sequence this was supposed to fix.

Pushed f780479 for the write amplification one: the KV write is now skipped whenever the poll repeats the previous result, only an actual transition writes. Added a test that reads the KV entry back and checks its version does not move across repeated polls but does increment on a real transition, so the skip is actually proven rather than just claimed. Same fix went to the .NET PR since you raised it here first.

Left the other two as you suggested:

- the `shouldEmit`/try-catch asymmetry: agreed a KV failure should not be silently absorbed the way a container failure is. If the store cannot be read or written, that is closer to an infra problem than "the script did not match" and I would rather the scheduler see the exception and retry than have the trigger quietly assume "not matched" and potentially miss a real transition.
- no TTL on the entry: I looked at `KVMetadata`, it does support one, but the safe value depends on the poll interval, which is user-configured and can be anything. Get it wrong and the entry expires while the flow is still polling, which reintroduces the exact bug this PR fixes. I would rather leave it unbounded than guess.

Malaydewangan09:
`f780479` is good. 34/34 pass here.

The thing I wanted to be sure the write-skip hadn't broken is the transition detection itself, since state is now only persisted on a change. Ran a full cycle in both directions through `evaluate()` with real containers, trigger reserialized every poll, stable context:

```
cmds : exit 1  exit 1  exit 0  exit 0  exit 1
got  : FIRED   quiet   quiet   quiet   FIRED
```

Both edges land where they should, so the optimization is behaviour preserving. Reading the version back rather than asserting on a mock is the right way to have tested it.

Agreed on leaving the other two, and your reasoning is better than my original note on both:

- the KV failure case, you're right that there's no safe default. Swallowing it means picking between a spurious fire and a missed transition, and neither is better than letting the scheduler see it and retry.
- the TTL, agreed. Any value has to exceed the poll interval, that's user configured and unbounded, and an entry expiring under an active flow puts back exactly the bug this fixes. Unbounded is the correct call. The orphaned-entry cleanup is a Kestra-wide housekeeping question, not something this PR should invent.

Nothing left from me on this one.

Malaydewangan09 (approved):

PR #442, opened 2026-09-18
fix(python): resolve HOME in an OS-independent way so tasks run on Windows

### LINKED ISSUE #441: Python tasks fail with NullPointerException on Windows: HOME is not defined
## What happens

On Windows, every `io.kestra.plugin.scripts.python.Script` task fails before `beforeCommands` or the script are reached:

```
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.toString()" because "<parameter2>" is null
	at java.base/java.lang.String.replace(Unknown Source)
	at io.kestra.plugin.scripts.python.internals.PythonDependenciesResolver.getUvCmd(PythonDependenciesResolver.java:339)
	at io.kestra.plugin.scripts.python.internals.PythonDependenciesResolver.findPython(PythonDependenciesResolver.java:566)
	at io.kestra.plugin.scripts.python.internals.PythonDependenciesResolver.findLocalPythonVersion(PythonDependenciesResolver.java:152)
	at io.kestra.plugin.scripts.python.internals.PythonEnvironmentManager.setup(PythonEnvironmentManager.java:79)
	at io.kestra.plugin.scripts.python.Script.run(Script.java:358)
```

## Cause

`PythonDependenciesResolver.java:36`:

```java
private static final String HOME_ENV = System.getenv("HOME");
```

**Windows does not define `HOME`** (it has `USERPROFILE`), so `HOME_ENV` is `null`. Two consequences:

1. **Line 339** — `Optional.ofNullable(System.getenv("UV_PATH")).orElse("$HOME/.local/bin/uv".replace("$HOME", HOME_ENV))` → `String.replace(target, null)`. Because `orElse` evaluates its argument eagerly, this throws **even when `UV_PATH` is set**, so there is no way to configure around it.
2. **Lines 384, 540, 577** — `env.put("HOME", HOME_ENV)` on a `ProcessBuilder` environment. Windows' `ProcessEnvironment` rejects a null value, which surfaces as a second NPE: `Cannot invoke "String.indexOf(int)" because "<parameter1>" is null`.

Code is identical in `v1.9.8` and `v1.9.10`.

## Reproduction

Kestra 1.3.37 standalone on Windows Server 2022, `plugin-script-python` 1.9.8, Process task runner:

```yaml
id: python-probe
namespace: test
tasks:
  - id: hello
    taskRunner:
    script: |
      print("hello")
```

Fails the same way with no Python installed, with system Python 3.12 + uv 0.12.15 on `PATH`, with a uv-managed interpreter and `UV_PYTHON_INSTALL_DIR` set, and with `pythonVersion` named explicitly.

## Confirmation

Setting `HOME` in the server's environment removes the NPE, and the resolver then works correctly on Windows:

```
uv --version                                 → uv 0.12.15 (x86_64-pc-windows-msvc)
uv python find --system --no-managed-python  → C:\Python312\python.exe
C:\Python312\python.exe --version            → Python 3.12.10
```

(The task then needs `interpreter: ["cmd", "/c"]`, since `AbstractExecScript` defaults to `/bin/sh -c` — a separate matter — after which it runs: `cmd /c python3 …py`, exit 0.)

## Suggested fix

Resolve the home directory portably — `HOME`, then `USERPROFILE`, then `System.getProperty("user.home")`, which the JVM always sets — and make line 339 use `orElseGet` so the default path is not computed when `UV_PATH` is set. Happy to open a PR.

ulise:
Fixes #441.

## Problem

`PythonDependenciesResolver` reads the home directory straight from the environment:

```java
private static final String HOME_ENV = System.getenv("HOME");
```

**Windows does not define `HOME`** — it uses `USERPROFILE` — so the field is `null` there and every Python task fails before `beforeCommands` or the script are reached:

```
java.lang.NullPointerException: Cannot invoke "CharSequence.toString()" because "<parameter2>" is null
	at java.lang.String.replace
	at PythonDependenciesResolver.getUvCmd(PythonDependenciesResolver.java:339)
	at PythonDependenciesResolver.findPython(:566)
	at PythonEnvironmentManager.setup(:79)
```

It bites in two places:

1. **`getUvCmd()`** builds the default uv path with `"$HOME/.local/bin/uv".replace("$HOME", HOME_ENV)`. Since `orElse` evaluates eagerly, this throws **even when `UV_PATH` is set** — so the value cannot be configured around.
2. **`env.put("HOME", HOME_ENV)`** at three `ProcessBuilder` call sites. Windows' `ProcessEnvironment` rejects null values, surfacing as a second NPE: `Cannot invoke "String.indexOf(int)" because "<parameter1>" is null`.

## Change

- Resolve the home directory as `HOME` → `USERPROFILE` → the `user.home` system property, which the JVM always sets, so the field can no longer be null on any platform. The lookup is injected as a `Function<String, String>` purely so it can be unit-tested.
- `getUvCmd()` uses `orElseGet`, so the default path is not computed when `UV_PATH` is set.

No behaviour change on Linux or macOS, where `HOME` is defined and is still preferred.

## Tests

Four unit tests on the new helper: `HOME` wins when set, `USERPROFILE` is used when it is not (the Windows case), `user.home` is the fallback, and a blank value never leaks through.

## Verification on a real host

Windows Server 2022, Kestra 1.3.37 standalone (Process task runner), `plugin-script-python` 1.9.8, Python 3.12.10 and uv 0.12.15. Setting `HOME` externally — the equivalent of this change — removes the NPE, and the resolver then completes correctly:

```
uv --version                                 → uv 0.12.15 (x86_64-pc-windows-msvc)
uv python find --system --no-managed-python  → C:\Python312\python.exe
C:\Python312\python.exe --version            → Python 3.12.10
```

after which a Python task runs to `exit code 0`.

One note for anyone testing on Windows afterwards: the task also needs `interpreter: ["cmd", "/c"]`, because `AbstractExecScript` defaults to `["/bin/sh", "-c"]` and `targetOS` only affects errexit handling. That is a separate matter in kestra core and is not touched here.


plugin-script-python/src/main/java/io/kestra/plugin/scripts/python/internals/PythonDependenciesResolver.java
```diff
@@ -33,10 +33,31 @@
  */
 public class PythonDependenciesResolver {
 
-    private static final String HOME_ENV = System.getenv("HOME");
+    private static final String HOME_ENV = resolveHomeDirectory(System::getenv);
     private static final String PATH_ENV = System.getenv("PATH");
     private static final String WORKING_DIR_ADDITIONAL_PYTHON_LIB = ".kestra_additional_python_lib";
 
+    /**
```
fdelbrayelle, line 40: 🟢 [Guidelines] Helper methods interrupt field/javadoc grouping
Problem: `resolveHomeDirectory`/`firstNonBlank` are inserted between the plain field block (`HOME_ENV`/`PATH_ENV`/`WORKING_DIR_ADDITIONAL_PYTHON_LIB`) and the javadoc for `DEFAULT_UV_INSTALLER_VERSION`, splitting what was one continuous "fields + their doc" section and making the class harder to scan top-to-bottom.
Fix: Move the two static helpers below the field-declaration block, e.g. next to `getUvCmd()`/`detectInstalledUvVersion()` which they support, so the field group stays contiguous.

plugin-script-python/src/test/java/io/kestra/plugin/scripts/python/internals/PythonDependenciesResolverTest.java
```diff
@@ -31,6 +32,43 @@ class PythonDependenciesResolverTest {
 
     private static final String FAKE_INSTALLER_URL = "https://astral.sh/uv/1.2.3/install.sh";
 
+    @Test
+    void shouldResolveHomeFromHomeEnvWhenSet() {
+        Function<String, String> env = Map.of(
```
fdelbrayelle, line 37: 🟢 [Guidelines] Prefer `var` for local variables
Problem: `Function<String, String> env = Map.of(...)::get;` repeats the explicit type across all four new tests (here and at lines 49, 56, 66) where `var` would infer it cleanly, per "Use var for all local variables whenever it is legally possible" (Java Style & Local Variables standard).
Fix: `var env = Map.of(...)::get;` in each of the four new tests.

plugin-script-python/src/test/java/io/kestra/plugin/scripts/python/internals/PythonDependenciesResolverTest.java
```diff
@@ -31,6 +32,43 @@ class PythonDependenciesResolverTest {
...
+            "HOME", "/home/kestra",
+            "USERPROFILE", "C:\\Users\\kestra"
+        )::get;
+
+        assertThat(PythonDependenciesResolver.resolveHomeDirectory(env), is("/home/kestra"));
+    }
+
+    @Test
+    void shouldResolveHomeFromUserProfileOnWindowsWhereHomeIsNotDefined() {
+        // Windows does not define HOME; reading it alone yields null, which then fails as an NPE
```
fdelbrayelle, line 47: 🟢 [Guidelines] Two-line comment
Problem: The comment above `shouldResolveHomeFromUserProfileOnWindowsWhereHomeIsNotDefined` spans two lines; the standard caps explanatory comments at "one short line max" outside of non-obvious algorithm documentation.
Fix: Trim to a single line, e.g. `// Windows has no HOME env var, only USERPROFILE.`

fdelbrayelle (approved):
## Kestra Plugin Code Review

Implements exactly the fix suggested in #441: resolve `HOME` -> `USERPROFILE` -> `user.home`, and switch the default uv path computation to `orElseGet` so it is not evaluated when `UV_PATH` is set. Verified end-to-end on a real Windows host per the PR description. A repo-wide grep confirms `PythonDependenciesResolver` is the only call site reading `System.getenv("HOME")`, so the fix is complete and does not leave a sibling code path unguarded.

### Kestra Guidelines — 3 findings (all nits, non-blocking)
🟢 Helper-method placement splits the field/javadoc grouping in the main class.
🟢 New test locals could use `var` per the "Java Style & Local Variables" standard.
🟢 One new test comment spans two lines where the standard caps explanatory comments at one line.

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
`String.replace(CharSequence, CharSequence)` is used (not `replaceAll`), so no regex-injection risk from a Windows path containing backslashes. No secrets, no new dependencies, no process/shell argument built from user-controlled task properties — `HOME`/`USERPROFILE`/`user.home` all come from the worker's own environment, not from task input.

### Performance — ✅ No issues found
The new logic runs once at class-load time (static field initializer) and involves only cheap `Optional` chaining; no loops, no blocking I/O, no unbounded allocation introduced.

**Verdict:** APPROVE

